PR #3, opened 2015-06-17
Handle parent directory checks on Windows

snoyberg:


DanBurton:
:+1:

chrisdone:
Works for me! Good catch.

PR #6, opened 2015-10-08
Improve a few property definitions

mgsloan:
The properties for stripDir and parent used the name 'parent' as a
variable.

The properties for filename and dirname seemed to be wrong and also used
'parent' as a variable.

### DIFF
--- a/src/Path.hs
+++ b/src/Path.hs
@@ -236,7 +236,7 @@ toFilePath (Path l) = l
 --
 -- The following properties hold:
 --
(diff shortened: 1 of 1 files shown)


chrisdone:
Sweet, nice improvements, thanks for reviewing these properties.

--

A fun thing to try in the future would be to automatically prove these properties with HERMIT and add it as a test-suite or embed proofs next to the declarations with template-haskell. :smile:

chrisdone:
Updated the docs on Hackage.

PR #7, opened 2015-11-21
Various improvements

mrkkrp:
Thanks for this library, I use it everywhere now. Most of these edits are completely optional, but type-safe synonyms for `toFilePath` is something I *always* define, simply because I want type-checker “double-check” me when I leave domain of this library and convert things into `FilePath` (which is just a `String` after all). Rationale is: you always know what kind of path you want to convert to `FilePath`, but `toFilePath` will convert *anything* into `FilePath`, even if it happens to be of wrong type (you think about absolute path, but it's relative — not impossible situation unless you put type-signatures for all intermediate transformations), so I don't trust myself, I trust type-checker, hence these synonyms.

----

There are quite a few commits there, how about publishing new release of the library? Thanks for your work.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -7,3 +7,4 @@ cabal-dev/
 TAGS
 tags
 *.tag
+.stack-work/
--- a/src/Path.hs
+++ b/src/Path.hs
@@ -81,7 +81,7 @@ parseAbsDir :: MonadThrow m
 parseAbsDir filepath =
   if FilePath.isAbsolute filepath &&
      not (null (normalizeDir filepath)) &&
-     not (isPrefixOf "~/" filepath) &&
+     not ("~/" `isPrefixOf` filepath) &&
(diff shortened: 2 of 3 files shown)


chrisdone:
I agree with these additional more type-safe functions, they are good practice and in line with the other functions in this library, i.e. always be specific about the type in contrast to being general which allows in mistakes. I'd be inclined to remove `toFilePath` if I were designing from the start, but that'd probably be an annoying change for others. So for now I will merge these in as-is, make a new release, and instigate some separate discussion about deprecating `toFilePath`, perhaps only in documentation rather than literally, because I think these functions would be useful which don't lose any expressivity.

:+1:

chrisdone:
Bumped to Hackage as 0.5.3. Thanks, @mrkkrp!

Pinging @borsboom who might be interested in these new functions.

PR #8, opened 2016-01-07
Adding a few tests that fail but shouldn't.

magthe:
These tests fail because the current implementation doesn't consider
paths containing `../` anywhere to be relative paths. I think that's
wrong, a relative path should be able to *start* with `../`.

### DIFF
--- a/test/Main.hs
+++ b/test/Main.hs
@@ -149,13 +149,18 @@ parseRelDirSpec =
      failing "//"
      failing "~/"
      failing "/"
-     failing "./"
      failing "//"
+     failing "./foo/../bar/"
+     failing "../foo/../bar/"
      failing "///foo//bar//mu/"
      failing "///foo//bar////mu"
      failing "///foo//bar/.//mu"
+     succeeding "./" (Path "./")
+     succeeding "../" (Path "../")
      succeeding "foo.bak" (Path "foo.bak/")
      succeeding "./foo" (Path "foo/")
+     succeeding "../foo" (Path "../foo/")
+     succeeding "./../foo" (Path "../foo/")
      succeeding "foo//bar//mu//" (Path "foo/bar/mu/")
      succeeding "foo//bar////mu" (Path "foo/bar/mu/")
      succeeding "foo//bar/.//mu" (Path "foo/bar/mu/")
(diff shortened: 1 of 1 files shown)


chrisdone:
`..` is intentionally rejected everywhere as something that doesn't compose. You can collapse your paths first like [here](https://github.com/commercialhaskell/stack/blob/master/src/Path/Extra.hs#L21). It might be nice to add `collapseFilePath` to the `filepath` package which is intended to be used with `path`.

chrisdone:
@ndmitchell Would you be in favor of adding such a `collapse` function to `filepath` if I made a self-contained function in a PR?

ndmitchell:
To replace `bar/../foo` with `foo`? Not really, since it's not correct in the presence of symlinks: http://neilmitchell.blogspot.co.uk/2015/10/filepaths-are-subtle-symlinks-are-hard.html . Or did you mean something different?

Note that `filepath` considers `../foo` to be a relative path.

ndmitchell:
For info, I am unlikely to accept any pull requests between now and GHC 8.0 final because it causes additional headaches for getting GHC released (but can leave pull requests in the queue until then).

chrisdone:
Fair enough.

magthe:
The issue as I see it is that a relative path should be able to *start* with `../` (as allowed by `filepath`).  I'm perfectly fine with not collapsing internal or trailing `../`.  I was just *extremely* surprised by stack's error message telling me that `../my-work-dir` wasn't a relative path when passed to `--work-dir`!

I do realise that composability of paths will suffer, but maybe relative paths simply aren't composable.

As it is now the function should probably be named `parseRelChildDir` and the error `InvalidRelChildDir` (only slightly tongue in cheek here).

chrisdone:
In that case it's the `--work-dir` parser that needs to be updated to use [the same one that's used for the `.yaml` file](https://github.com/fpco/stack/blob/master/src/Path/IO.hs#L69..L70), which supports `../`.

> I do realise that composability of paths will suffer, but maybe relative paths simply aren't composable.

They are as implemented in `path`. I'd prefer to leave it out. Supporting `../` and `~/` and `./` and such are classed, by me, as UI issues.

chrisdone:
I think just [here](https://github.com/fpco/stack/blob/master/src/Stack/Config.hs#L104..L104) needs to be updated to use `parseRelAsAbsDir`.

magthe:
Thanks, I'll have a look at fixing it in stack then.  I'll also file a PR with documentation of the limitations of relative dirs in `path`.

chrisdone:
It does currently state in the path docs that `..` is disallowed completely.

magthe:
That's for the data type, I think it would be clearer to have a comment on that for the *parsers*.

magthe:
There are good reasons for the `path`-flavour of relative paths, so this PR can be closed.

PR #9, opened 2016-01-07
Clarify what limitations the RelX parsers have.

magthe:
The behaviour is surprising given the rather established meaning of *relative path*.  Since `path` doesn't quite follow that meaning (and for good reasons) the next best thing is clarifying documentation.


magthe:
Yupp, that would be the "good reasons" I refer to.

PR #12, opened 2016-01-28
Improve description of the package

mrkkrp:
See comment of the second commit.

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,4 +1,4 @@
-Copyright (c) 2015, FP Complete
+Copyright (c) 2015–2016, FP Complete
 All rights reserved.
 
 Redistribution and use in source and binary forms, with or without
--- a/README.md
+++ b/README.md
@@ -1,4 +1,518 @@
-path
-=====
+# Path
 
 Support for well-typed paths in Haskell.
+
+* [Motivation](#motivation)
+* [Approach](#approach)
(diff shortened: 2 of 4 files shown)


mrkkrp:
Can you enable issues for this repo?

chrisdone:
Nice work, cheers!

mrkkrp:
Oh, I just have a snippet for these things and `experimental` is most commonly used option for new packages and I left it as is.

mrkkrp:
I would like to ask you again to enable issues for this repo. Since I cannot open an issue, I will point out right here that this:

```
λ> parseRelFile "."
"."
```

Is probably wrong, this should not succeed. On Unix-like systems you cannot have a file named `"."`, because this name is already assigned to current directory, thus this cannot be valid relative file name.

mrkkrp:
@chrisdone, Also, why not to use `isValid` checks provided by `filepath` when parsing paths?

chrisdone:
@mrkkrp Good catch, `parseRelFile` should reject `"."`. I'm going to fix that. I've enabled issues on the repo.

I'll check out `isValid` to see what it gets us.

chrisdone:
@mrkkrp It seems that `isValid` buys us some checks. I've included use of it as `0.5.5`. Thanks for the suggestion!

PR #29, opened 2016-06-28
Update documentation of parsing functions.

harendra-kumar:
Clarified the semantics of a valid or expected path.


chrisdone:
Nice!

PR #38, opened 2016-09-03
Stricter validity instance that exposes some bugs

NorfairKing:
This was made as a reply to #37.

This PR makes the validity instances for typed path more strict and exposes some bugs in doing so:

```
$(mkAbsFile "/.") -- DANGEROUS
"/"
$(mkRelFile "a/.") -- Disobeys invariant: file does not end in slash
"a/"
$(mkRelFile "a..") -- Disobeys invariant: no double dot in filepath
"a.."
$(mkRelDir "a../") -- Disobeys invariant: no double dot in filepath
"a../"
```

Fixes are not part of this PR.

### DIFF
--- a/test/Path/Gen.hs
+++ b/test/Path/Gen.hs
@@ -14,30 +14,33 @@ import           Test.QuickCheck
 
 
 instance Validity (Path Abs File) where
-  isValid (Path fp)
+  isValid p@(Path fp)
     =  FilePath.isAbsolute fp
     && not (FilePath.hasTrailingPathSeparator fp)
     && FilePath.isValid fp
     && not (".." `isInfixOf` fp)
+    && (parseAbsFile fp == Just p)
 
 instance Validity (Path Rel File) where
-  isValid (Path fp)
+  isValid p@(Path fp)
     =  FilePath.isRelative fp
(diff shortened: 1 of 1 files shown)


NorfairKing:
Any progress here?

chrisdone:
Okay, I've fixed the abs file `/.` case, and the `a/.` rel file case. The other two I'm not sure are invalid.

```
Chriss-MacBook-Pro:~ chris$ touch a..
Chriss-MacBook-Pro:~ chris$ echo hi > a..
Chriss-MacBook-Pro:~ chris$ cat a..
hi
Chriss-MacBook-Pro:~ chris$ mkdir foo..
Chriss-MacBook-Pro:~ chris$ ls foo..
Chriss-MacBook-Pro:~ chris$ touch foo../blah
Chriss-MacBook-Pro:~ chris$ ls foo..
blah

```

chrisdone:
Pushed as 0.5.10.

NorfairKing:
Thanks!

It's okay to consider `a..` valid but then the invariant has to go, or need to be modified to allow for it.

NorfairKing:
I will add some more tests for other properties soon.

PR #40, opened 2016-10-31
add trivial 'isRoot' helper

emmanueltouzery:
didn't add tests but if 'parent' is correct then 'isRoot' must be...


mrkkrp:
This relies on behavior that is not good, see #18 and #33. Closing. I'll add an issue requesting this feature so we don't forget about it.

PR #42, opened 2016-11-25
Add ‘fileExtension’ and ‘setFileExtension’

mrkkrp:
Allow to get and set file extension. I'm tired of converting to and form `FilePath` every time I need this.

I tried to follow your naming style, but well, `filext` may be not the best name, let me know if you have a better naming idea.

### DIFF
--- a/src/Path.hs
+++ b/src/Path.hs
@@ -40,6 +40,8 @@ module Path
   ,parent
   ,filename
   ,dirname
+  ,fileExtension
+  ,setFileExtension
   -- * Conversion
   ,toFilePath
   ,fromAbsDir
@@ -53,12 +55,14 @@ import           Control.Exception (Exception)
 import           Control.Monad.Catch (MonadThrow(..))
 import           Data.Aeson (FromJSON (..))
 import qualified Data.Aeson.Types as Aeson
+import           Data.Char (isAlphaNum)
 import           Data.Data
 import           Data.List
 import           Data.Maybe
 import           Language.Haskell.TH
 import           Path.Internal
 import qualified System.FilePath as FilePath
+import           Unsafe.Coerce (unsafeCoerce)
 
 --------------------------------------------------------------------------------
 -- Types
@@ -110,6 +114,7 @@ data PathParseException
   | InvalidAbsFile FilePath
   | InvalidRelFile FilePath
   | Couldn'tStripPrefixDir FilePath FilePath
+  | InvalidFileExtension String
   deriving (Show,Typeable)
 instance Exception PathParseException
 
@@ -148,7 +153,8 @@ parseRelDir filepath =
   if not (FilePath.isAbsolute filepath) &&
      not (hasParentDir filepath) &&
      not (null filepath) &&
-     filepath /= "." && (normalizeFilePath filepath) /= curDirNormalizedFP &&
+     filepath /= "." &&
+     normalizeFilePath filepath /= curDirNormalizedFP &&
      filepath /= ".." &&
      FilePath.isValid filepath
      then return (Path (normalizeDir filepath))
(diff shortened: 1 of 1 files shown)


mrkkrp:
I wonder what happens though when you try to `setFilext "something/slashy"`… Need to check.

mrkkrp:
I think we must check extension for validity so it's not anything that may invalidate given path. Adding a new data constructor to `PathParseExtension` and makng `setFilext` live in a `MonadThrow` may be a good idea.

mrkkrp:
The failing build is not related to the change.

----

Another idea is to add `Extension` type with similar smart constructors and TH helpers and use that as argument. Then we don't need to make `setFilext`  living a `MonadThrow`. Maybe it's a bit heavy-weight solution though.

mrkkrp:
@chrisdone First mergeable version is there, please take a look. I can change the check if necessary and add tests.

mrkkrp:
Ping?

chrisdone:
Apart from the `unsafeCoerce` this looks fine.

mrkkrp:
@chrisdone This use of `unsafeCoerce` is safe AFAIK. A better idea how to make it work without it/reason we should not use it here?

mrkkrp:
It just converts between two `newtype` wrappers around the same `String` type.

chrisdone:
I've fixed the commit [here](https://github.com/chrisdone/path/commit/b135ac589adbd094e62f991a1aab444731b6083e).

This was a pretty poor PR, @mrkkrp. You really have to test out your code:

```
bash-3.2$ pwd
/Users/chris/Work/path
bash-3.2$ git checkout 4aea22c04f96f897867154c71a7caaaf15539080
Note: checking out '4aea22c04f96f897867154c71a7caaaf15539080'.
HEAD is now at 4aea22c... Add ‘fileExtension’ and ‘setFileExtension’
bash-3.2$ stack ghci
Prelude Path Path.Internal> setFileExtension "hs" $(mkAbsFile "/foo/bar.txt")
<interactive>: internal error: stg_ap_v_ret
    (GHC version 7.10.3 for x86_64_apple_darwin)
    Please report this as a GHC bug:  http://www.haskell.org/ghc/reportabug
Abort trap: 6
```

Even correcting the `unsafeCoerce` with a `coerce`, the arguments to `replaceExtension` are the wrong way round:

```haskell
Prelude Path Path.Internal> setFileExtension "hs" $(mkAbsFile "/foo/bar.txt")
"hs./foo/bar.txt"
```

Be more careful next time! 😅 👍

chrisdone:
> This use of unsafeCoerce is safe AFAIK. A better idea how to make it work without it/reason we should not use it here? It just converts between two newtype wrappers (with different phantom types) around the same String type.

In this case we can use the recent roles-based `Data.Coerce` function, which only performs safe coercions of data types. So the code ends up like:

```haskell
setFileExtension :: MonadThrow m
  => String            -- ^ Extension to set
  -> Path b File       -- ^ Old file name
  -> m (Path b File)   -- ^ New file name with the desired extension
setFileExtension ext (Path path) =
  if FilePath.isAbsolute path
    then fmap coerce (parseAbsFile (FilePath.replaceExtension path ext))
    else fmap coerce (parseRelFile (FilePath.replaceExtension path ext))
```

Where `fmap coerce` simply coerces the `Abs` or `Rel` type into the more general `b`.

mrkkrp:
Uh oh, indeed! (This is was rather a sketch, I didn't have much time.) Don't forget to drop `InvalidFileExtension` now.

mrkkrp:
Ah I forgot to add `@since` declarations to the docs. Can you do it now?

chrisdone:
@mrkkrp What's the syntax like for that?

mrkkrp:
https://hackage.haskell.org/package/base-4.9.0.0/docs/src/Data.Bool.html#bool

See docs for `bool`, it has `@since 4.7.0.0`.

chrisdone:
Added `@since` to both and bumped as 0.5.11.

PR #43, opened 2016-12-08
Property tests

NorfairKing:
This property is advertised:
`filename (p </> a) == filename a`, but it was tested wrong, and only with a unit test.
This PR fixes that.


NorfairKing:
Accidental formatting changes.

PR #45, opened 2016-12-12
Add ‘Hashable’ instance for ‘Path’

mrkkrp:


chrisdone:
I'd like to add the hashable instance without the Generic instance which breaks abstraction boundaries.

mrkkrp:
Dropped generics.

chrisdone:
Tip top! Thanks.

PR #58, opened 2017-02-08
Docs for `show` are misleading

3noch:
Using `show` will include double quotes (as usual when `show`ing a `String`) but docs make it sound like the quotes are not included.


chrisdone:
Good catch.

PR #96, opened 2017-04-24
restrict kinds for Path data type

queertypes:
## Overview

This change makes it more clear at the API level what sorts of types
can be assigned to the Path data type. Furthermore, it makes error
messages more helpful when there's a kind mismatch.

Further, two new data types are introduced: PathRelativity and
FileType. These sum types make it clear that Abs | Rel are
mutually-exclusive, and similarly for File | Dir.

Tests and sources have been updated accordingly, including enabling
DataKinds and KindSignatures where needed.

## Detailed Summary

The changes are essentially, going from:

```haskell
data Rel deriving (Typeable)
data Abs deriving (Typeable)
data File deriving (Typeable)
data Dir deriving (Typeable)
newtype Path b t = Path FilePath
```

to

```haskell
{-# LANGUAGE DataKinds #-}
{-# LANGUAGE KindSignatures #-}

data PathRelativity = Abs | Rel deriving (Typeable)
data FileType = File | Dir deriving (Typeable)
newtype Path (b :: PathRelativity) (t :: FileType) = Path FilePath
```

## Final Notes

All tests pass on Linux. I've modified the Windows tests as need, but I haven't been able to test for myself.

The public API **should** be identical, even though some things needed to be moved around.

Let me know if y'all are interested in this PR.

### DIFF
--- a/src/Path.hs
+++ b/src/Path.hs
@@ -6,14 +6,13 @@
 {-# LANGUAGE DeriveDataTypeable #-}
 {-# LANGUAGE EmptyDataDecls #-}
 {-# LANGUAGE FlexibleInstances #-}
+{-# LANGUAGE DataKinds #-}
 
 module Path
   (-- * Types
    Path
-  ,Abs
-  ,Rel
-  ,File
-  ,Dir
+  ,PathRelativity(..)
+  ,FileType(..)
   -- * QuasiQuoters
   -- | Using the following requires the QuasiQuotes language extension.
   --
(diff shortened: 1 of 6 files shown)


joehillen:
> Furthermore, it makes error
messages more helpful when there's a kind mismatch.

Can you show an example of how the error messages improve? Will this work with older GHCs?

queertypes:
Should work for GHC>=7.4.1, according to the GHC user's guide. specifically​, DataKinds & KindSignatures were added then.

After I get some sleep, I'll generate some error messages for before & after and share them on here.

Thanks!

NorfairKing:
> Furthermore, it makes error
> messages more helpful when there's a kind mismatch.

I was wondering if there was anything that could fix that.

If I understand correctly the tradeoffs are as follows:

pros: 
- Better error messages
- Makes the assumptions of this library explicit
cons:
- people can't define their own relativity or filetype anymore

I like this tradeoff.
I'm not in favor of the name `FileType`, but I like the overall proposal.

Please do make sure that the tests pass though.

On a related note. 
Kan we split this proposal into two pieces: One where the arguments must be of kind `*`, and the second where they have to be these specific data kinds?

mrkkrp:
I think @chrisdone explicitely states in readme why he has chosen the current approach. Do you think Chris was unaware of possibility of this design?

This is also not backward-compatible, as it requires to write `Path 'Abs 'Dir` instead of `Path Abs Dir`. I'm not in favor of merging this.

chrisdone:
I wrote about that here: http://chrisdone.com/posts/path-package#implementation

> Why not use data kinds like `data Type = File | Dir`? Because that imposes an extension overhead of adding `{-# LANGUAGE DataKinds #-}` to every module you might want to write out a path type in. Given that one cannot construct paths of types other than these, via the operations in the module, it’s not a concern for me.

It seems like a not-worth-it patch to me. An example of error messages improved would still be interesting, though, as that appears to be the main motivation here.

queertypes:
I'll go ahead and close this out then.

@mrkkrp, I do not presume to predict what design spaces Chris has explored or considered. Your commentary was unhelpful.

@chrisdone that makes sense! Thanks for linking.

PR #97, opened 2017-06-07
Add a synonym for ‘setFileExtension’, as ‘(-<.>)’

mrkkrp:


NorfairKing:
I like the idea, but there should probably be doctests, because it doesn't do what you think it does. (same for `setFileExtension`).
Up to you if you want these examples in this PR.

I'm +1 on this one in any case.

NorfairKing:
Please tell me if I'm wrong, but as far as I remember it removes all extensions and sets a new one.

So `file.tar.gz -<.> txt` is `file.txt` instead of `file.tar.txt`.
It's good that this is what it does, but it's not what I expected.

mrkkrp:
The `(-<.>)` is expected to work like `replaceExtension` from `filepath` (which also has `(-<.>)` with exactly the same behavior). Maybe you're confusing it with `(<.>)` which indeed adds an extension. The name `setFileExtension` and the associated Haddock also indicate quite clearly what it does, although maybe we should steal some examples from `filepath` and add them there.

NorfairKing:
Maybe I shouldn't have started this discussion here. I like this PR. Feel free to merge :)

mrkkrp:
I think I'll add some tests for `setFileExtension` first. Strange, I thought we had this thing tested...

NorfairKing:
Both of the tests you added could be property tests...

mrkkrp:
If we are to switch to property tests in the test suite, we should start with "upgrading" all the existing tests. Out of scope for this PR.

NorfairKing:
You could just add it here: https://github.com/commercialhaskell/path/blob/master/test/ValidityTest.hs

PR #101, opened 2017-06-13
Rename, export, document exceptions

harendra-kumar:
Also:

1) Rename PathParseException and one of its constructors
2) Export the constructors so that info about exceptions can be extracted.
3) Document which specific exception is thrown during each operation.

These are the remaining changes from #35 . These are breaking so should go before the next major version release.

### DIFF
--- a/src/Path.hs
+++ b/src/Path.hs
@@ -25,6 +25,8 @@ module Path
   ,Rel
   ,File
   ,Dir
+   -- * Exceptions
+  ,PathException(..)
   -- * QuasiQuoters
   -- | Using the following requires the QuasiQuotes language extension.
   --
@@ -54,7 +56,6 @@ module Path
   ,parseRelDir
   ,parseAbsFile
   ,parseRelFile
-  ,PathParseException
   -- * Conversion
   ,toFilePath
   ,fromAbsDir
@@ -68,6 +69,7 @@ module Path
   ,mkAbsFile
   ,mkRelFile
   -- * Deprecated
+  ,PathParseException
   ,stripDir
   ,isParentOf
   )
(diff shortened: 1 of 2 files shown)


harendra-kumar:
Can someone tell me why the cabal tests are failing?

src/Path.hs
```diff
@@ -116,16 +118,15 @@ parseJSONWith f x =
        Left e -> fail (show e)
 {-# INLINE parseJSONWith #-}
 
--- | Exception when parsing a location.
-data PathParseException
+-- | Exceptions that can occur during path operations
```
mrkkrp: Can we end the sentence with a full stop? (Sorry for nitpicking!)

src/Path.hs
```diff
@@ -520,6 +523,9 @@ normalizeFilePath = normalizeLeadingSeparators . FilePath.normalise
 --------------------------------------------------------------------------------
 -- Deprecated
 
+{-# DEPRECATED PathParseException "Please use PathException instead." #-}
+type PathParseException = PathException
```
mrkkrp: Add doc-string to this one and Cabal builds should pass. We catch not-100% Haddock coverage, not sure why it's so unfriendly though (my typical setup prints per-module coverages).

mrkkrp:
@harendra-kumar Do you want to make a write-up about all the changes in 0.6.0? I think it makes sense to write a blog post or something to highlight the changes.

harendra-kumar:
@mrkkrp I have been thinking of starting a blog post but have been too lazy. It might take me a couple of weeks to start one. You can write a blog post if you feel like it, I will be happy to review it.

mrkkrp:
@harendra-kumar Here is the blog post: https://markkarpov.com/post/path-0.6.0-released.html. Can you take a look before I announce the post on social media?

harendra-kumar:
@mrkkrp the post looks good. I would suggest the following edit (you can edit it to your taste):

> Another change is related to the parent function. Previously it only could be used with absolute 
> paths, now you can get parent of a relative path as well:

Having `.` as a relative directory path also allows `parent` to work on relative paths. `parent` now has the following semantics:

```
parent (x </> y) == x
parent "/x" == "/"
parent "x" == "."
parent "/" = "/"
parent "." = "."
```

In general, addition of `.` as a relative path makes relative path APIs behave quite similar to absolute path APIs in all cases where `.` behaves as a relative dual of `/`.

PR #104, opened 2017-06-16
Fix the documentation for "parent", "dirname" and "</>"

harendra-kumar:
1) consistent format with other existing usage (use $(mkRelDir "x") instead of "x")
2) add examples of more cases for "."

### DIFF
--- a/src/Path.hs
+++ b/src/Path.hs
@@ -231,6 +231,14 @@ relfile = qq mkRelFile
 --
 -- @x \<\/> $(mkAbsDir …)@
 --
+-- The following equalities hold for the special relative directory @.@:
+--
+-- @x \<\/> $(mkRelDir ".") == x@
+--
+-- @$(mkRelDir ".") \<\/> x == x@
+--
+-- @$(mkRelDir ".") \<\/>  $(mkRelDir ".") == $(mkRelDir ".")@
+--
 infixr 5 </>
 (</>) :: Path b Dir -> Path Rel t -> Path b t
 (</>) (Path a) (Path b) = Path (a ++ b)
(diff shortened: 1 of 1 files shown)


src/Path.hs
```diff
@@ -315,9 +323,18 @@ filename (Path l) =
...
+--
 -- @dirname $(mkRelDir ".") == $(mkRelDir ".")@
 --
--- @dirname (p \<\/> a) == dirname a@
+-- @dirname $(mkRelDir "x") == $(mkRelDir "x")@
+--
+-- The following POSIX path properties (and equivalent Windows path properties)
+-- hold:
+--
+-- @dirname $(mkAbsDir "/") == $(mkRelDir ".")@
```
joehillen: Does this `/` need to be escaped?

src/Path.hs
```diff
@@ -231,6 +231,14 @@ relfile = qq mkRelFile
 --
 -- @x \<\/> $(mkAbsDir …)@
 --
+-- The following equalities hold for the special relative directory @.@:
+--
+-- @x \<\/> $(mkRelDir ".") == x@
```
joehillen: We should probably include some QuasiQuotes examples too

```
x </> [reldir|.]
```

src/Path.hs
```diff
@@ -279,18 +287,18 @@ isProperPrefixOf p l = isJust (stripProperPrefix p l)
...
 -- The following properties hold:
 --
--- @
--- parent (x \<\/> y) == x
--- parent \"\/x\" == \"\/\"
--- parent \"x\" == \".\"
--- @
+-- @parent (x \<\/> y) == x@
 --
--- On the root (absolute or relative), getting the parent is idempotent:
```
joehillen: Why remove this comment? I think it should stay.

src/Path.hs
harendra-kumar: No, the core point here is to illustrate the property. Either TH splice or quasiquote should be enough. And whichever we chose should be used everywhere consistently.

src/Path.hs
harendra-kumar: Two reasons:

1) To keep that comment I have to group the root related properties together, instead I wanted to group relative and absolute properties separately. It also makes the `parent` and `dirname` properties format the same.

2) It is obvious from the rules.

src/Path.hs
harendra-kumar: No, it works fine.

harendra-kumar:
Languished for too long. I am closing this.

Issue #105, opened 2017-06-24
Add `addExtension` analogue from `filepath`


ChShersh:
Sometimes `setExtension` (or `(-<.>)` operator) is not enough. Especially when working with Haskell-like files:

```haskell
λ: hsFileName <- parseRelFile "Data.List"
λ: hsFileName 
"Data.List"
λ: setFileExtension ".symbols" hsFileName
"Data.symbols"  -- want to have Data.List.symbols here
```

I propose to add `addExtension` function similar to [one from `filepath` library](http://hackage.haskell.org/package/filepath-1.4.1.2/docs/System-FilePath-Posix.html#v:addExtension) with corresponding `(<.>)` operator. It may have next signature similar to `setExtension`:

```haskell
addExtension :: MonadThrow m
  => String            -- ^ Extension to add
  -> Path b File       -- ^ Old file name
  -> m (Path b File)   -- ^ New file name with the desired extension added at the end
```

mrkkrp:
I'm in favor.

NorfairKing:
Me too.

joehillen:
I'm in favour of this as well.

@ChShersh would you be willing your write a PR for this?

ChShersh:
@joehillen I don't mind implementing this addition as a PR. Will do this relatively soon.

PR #106, opened 2017-06-27
Add `addFileExtension` function and `(<.>)` operator

ChShersh:
This PR add function that adds file extension according to #105. I'm not sure I did everything right because I didn't notice contribution policies. But I tried. Please, correct me, if I need to add/fix anything.

### DIFF
--- a/src/Path.hs
+++ b/src/Path.hs
@@ -335,8 +335,16 @@ fileExtension = FilePath.takeExtension . toFilePath
 -- | Add extension to given file path. Throws if the
 -- resulting filename does not parse.
 --
+-- >>> addFileExtension "txt $(mkRelFile "foo")
+-- "foo.txt"
 -- >>> addFileExtension "symbols" $(mkRelFile "Data.List")
 -- "Data.List.symbols"
+-- >>> addFileExtension ".symbols" $(mkRelFile "Data.List")
+-- "Data.List.symbols"
+-- >>> addFileExtension "symbols" $(mkRelFile "Data.List.")
+-- "Data.List..symbols"
+-- >>> addFileExtension ".symbols" $(mkRelFile "Data.List.")
+-- "Data.List..symbols"
 -- >>> addFileExtension "evil/" $(mkRelFile "Data.List")
 -- *** Exception: InvalidRelFile "Data.List.evil/"
 --
@@ -353,6 +361,7 @@ addFileExtension ext (Path path) =
         coercePath (Path a) = Path a
 
 -- | A synonym for 'addFileExtension' in the form of an operator.
+-- See more examples there.
 --
 -- >>> $(mkRelFile "Data.List") <.> "symbols"
 -- "Data.List.symbols"
--- a/src/Path.hs
+++ b/src/Path.hs
@@ -49,6 +49,8 @@ module Path
   ,filename
   ,dirname
   ,fileExtension
+  ,addFileExtension
+  ,(<.>)
   ,setFileExtension
   ,(-<.>)
    -- * Parsing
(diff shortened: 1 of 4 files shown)


NorfairKing:
I would also like some property tests: https://github.com/commercialhaskell/path/blob/master/test/ValidityTest.hs, but doesn't have to be in this PR.

src/Path.hs
```diff
@@ -330,6 +332,41 @@ dirname (Path l) = Path (last (FilePath.splitPath l))
...
+-- >>> addFileExtension "symbols" $(mkRelFile "Data.List")
+-- "Data.List.symbols"
+-- >>> addFileExtension "evil/" $(mkRelFile "Data.List")
+-- *** Exception: InvalidRelFile "Data.List.evil/"
+--
+-- @since 0.6.1
+addFileExtension :: MonadThrow m
+  => String            -- ^ Extension to add
+  -> Path b File       -- ^ Old file name
+  -> m (Path b File)   -- ^ New file name with the desired extension added at the end
```
joehillen: Why is the argument order different than in [filepath](https://hackage.haskell.org/package/filepath-1.4.1.2/docs/System-FilePath-Posix.html#v:addExtension)? I think I prefer filepath's version.

src/Path.hs
```diff
@@ -330,6 +332,41 @@ dirname (Path l) = Path (last (FilePath.splitPath l))
...
+-- | Add extension to given file path. Throws if the
+-- resulting filename does not parse.
+--
+-- >>> addFileExtension "symbols" $(mkRelFile "Data.List")
+-- "Data.List.symbols"
+-- >>> addFileExtension "evil/" $(mkRelFile "Data.List")
+-- *** Exception: InvalidRelFile "Data.List.evil/"
+--
+-- @since 0.6.1
+addFileExtension :: MonadThrow m
```
joehillen: Do we really need the word `File` in the function name? What is wrong with `addExtension`?

src/Path.hs
NorfairKing: I disagree. This is easier to map.

joehillen: Good point.

src/Path.hs
ChShersh: @joehillen I did it similar to `setFileExtension`. I thought consistency is better.

src/Path.hs
```diff
@@ -330,6 +332,41 @@ dirname (Path l) = Path (last (FilePath.splitPath l))
...
+addFileExtension ext (Path path) =
+  if FilePath.isAbsolute path
+    then liftM coercePath (parseAbsFile (FilePath.addExtension path ext))
+    else liftM coercePath (parseRelFile (FilePath.addExtension path ext))
+  where coercePath :: Path a b -> Path a' b'
+        coercePath (Path a) = Path a
+
+-- | A synonym for 'addFileExtension' in the form of an operator.
+--
+-- >>> $(mkRelFile "Data.List") <.> "symbols"
```
ChShersh: @joehillen Sure, I can add more examples into comments.

src/Path.hs
joehillen: Fair enough

src/Path.hs
ChShersh: @joehillen @mrkkrp Btw, should I make examples expandable how it's done for [`Data.Bool.bool`](http://hackage.haskell.org/package/base-4.9.1.0/docs/Data-Bool.html#v:bool)? There will be more than couple examples so probably it makes sense to hide them by default.

mrkkrp: Personally, I like how it's done here: https://hackage.haskell.org/package/filepath-1.4.1.2/docs/System-FilePath-Posix.html#v:splitExtension, you can see all the subtle details of behavior right away, and I don't think it wastes too much space either. Use your judgement though.

joehillen: I also don't like hiding examples and prefer them front and center. Haddock's CSS make hidden examples easy to miss.

ChShersh:
I've written examples not like in `filepath` library. I used `ghci`-like examples so later they can be easily verified and kept up-to-date using `doctest`. Probably adding `doctest` is a good idea and it's worth it.

joehillen:
LGTM. @magthe or @NorfairKing either of you can merge it if it looks good to you.

I agree about adding doctest. It's been on my mental TODO list for a while.

Issue #107, opened 2017-08-02
resolveDir is present in docs, but missing from the library


varosi:
In this documentation:
https://github.com/commercialhaskell/path

you'll find
```haskell
resolveDir :: (MonadIO m, MonadThrow m)
           => Path Abs Dir -> FilePath -> m (Path Abs Dir)
```
but it's missing from the library.

mrkkrp:
It's here: https://www.stackage.org/haddock/nightly-2017-07-31/path-io-1.3.3/Path-IO.html#v:resolveDir

varosi:
Thanks a lot! I have just found it, but didn't know that it's in a separate package.

PR #108, opened 2017-08-02
Added more info where to find resolveDir function.

varosi:
This was not obvious for me from the first read. So I hope that I could save some time of first-starters.

Related to https://github.com/commercialhaskell/path/issues/107


mrkkrp:
Maybe it's better to link to Hackage? I use Stackage to browse the docs and everything, but I think Hackage is still more "canonical" source when we talk about packages.

mrkkrp:
Thanks!

PR #113, opened 2018-01-14
Added basename support

varosi:
Before it was harder to add suffix to basename, like: image.jpg -> image_bw.jpg because lib were missing function to get base name, i.e. filename without extension.


joehillen:
I can't check right now, but how does this handle multiple extensions, like `foo.tar.gz`? Can you add test cases for that?

joehillen:
I can't check right now, but how does this handle multiple extensions, like `foo.tar.gz`? Can you add test cases for that?

mrkkrp:
As a general rule, please avoid changing styling in PRs that add some functionality. This makes it harder to see changes that really matter.

harendra-kumar:
I would like this to be called `fileBasename` to be consistent with other file API names and as a consistently named counterpart of `fileExtension`. The file prefix also tells you that this is a file operation and does not apply to a dir. Furthermore, it avoids confusion with the POSIX `basename` API which works for both files and directories.

NorfairKing:
I'd actually prefer it be called `dropExtension` and a seperate function for `dropExtensions` where the first drops one extension and the second drops all extensions.

joehillen:
@NorfairKing I like that idea best, but I'd worry about an insidious bug because of a unintentional dangling `s`. I'd prefer a more verbose name for `dropExtensions`, maybe `dropAllExtensions`?

It would be nice if we could just tell people to use the long form, but I'd don't think it would be obvious enough:

```haskell
> until (("" ==) . fileExtension) dropExtension [relfile|foo.tar.gz|]
"foo"
```

varosi:
@joehillen, yes, it handles multiple extensions. It's using [takeBaseName](http://hackage.haskell.org/package/filepath-1.4.1.2/docs/System-FilePath-Posix.html#v:takeBaseName)

varosi:
@mrkkrp, sorry, I'm using stylish-haskell for auto corrections and I didn't saw this change. Now I brought it back.

harendra-kumar:
@NorfairKing, @joehillen - given the semantics, `dropExtension` is not the same as basename. basename as implemented in this PR strips both the directory prefix and the extension e.g. the basename of `/x/y/z.ext` would be `z`. 

However we can just add the `dropExtension` API and use a combination of `filename` and `dropExtension` to achieve what `basename` is doing in this PR. I would prefer that because it is more modular.

varosi:
hm, it's not only dropping extension, but it is returning only base file name, i.e.:
dropExtension $(mkRelFile "b\\test.txt")
"test"

varosi:
Is it okay to bring it back to "fileBaseName" ?

harendra-kumar:
If we implement a real `dropExtension` we do not need a `fileBasename` as I suggested in my previous comment. To implement `dropExtension` we  just need to use the `System.Filepath.dropExtension` API and there is a `System.Filepath.dropExtensions` API as well if we want that.

harendra-kumar:
My point was that if we just implement `dropExtension` then `fileBasename` can be trivially derived as: `fileBasename = dropExtension . filename`.  We do not have to write it as a library function users can do it themselves whenever they need.

varosi:
Okay, do we all agree to have
```haskell
dropExtension :: Path b File -> Path b File
```
?

NorfairKing:
Yes, but also `dropAllExtensions`. But how do you define it for `.vimrc`?

harendra-kumar:
The `filepath` package defines it like this:

```
Prelude System.FilePath> dropExtension ".vimrc"
""
```

However, I think the correct result should be `.vimrc` because practically we need a file to have an extension so the first part should always be treated as a file. We can reason by flipping the question around - can we create `.vimrc` by adding an extension to some file? The answer is we cannot. That file will have to be "" and we do not allow creation of file with an empty name. So the only correct result of `dropExtension $(mkRelFile ".vimrc")` seems to be `.vimrc` itself.

NorfairKing:
@harendra-kumar you bumped into exactly the problem I wanted to bring up for discussion.

varosi:
I think we should use different type:
```haskell
dropExtension :: Path b File -> Maybe (Path b File)
```
because result of such paths bring empty base name which is not valid Path object.

harendra-kumar:
The semantics of this function are: if the file has an extension drop it otherwise leave it alone. The file ".vimrc" does NOT have an extension so we leave it alone. Note that ".vimrc" is not a file with an extension "vimrc" instead ".vimrc" is a filename without an extension.

varosi:
Is it alright for everyone to use Maybe?

varosi:
This is unpredictable behaviour! If I would like to add suffix and bring back the extension - it'll add second extension to the original one.

NorfairKing:
@varosi Please read @harendra-kumar's considerations closely again.
The empty string is not a valid filepath.

joehillen:
I don't know anything about Windows, but if files starting with dots are not valid files on Windows, then that should be prohibited during construction. Currently it's not:
```haskell
> [Path.Windows.relfile|.vimrc|]
".vimrc"
```
I'm also curious how Windows handles trailing dots:
```haskell
> [Path.Windows.relfile|.foo.|]
".foo."
```
or multiple trailing dots:
```haskell
> [Path.Windows.relfile|foo....|]
"foo...."
```
If this isn't valid, it should be prohibited, and we should fix that.

joehillen:
@varosi FYI, your validity tests failed: https://travis-ci.org/commercialhaskell/path/jobs/329227303#L887

harendra-kumar:
".vimrc" is a valid file on Windows as well. You can easily create it using "echo > .vimrc" on command prompt.

I could not create a file with trailing dots (single or multiple) on Windows from the command prompt. samba/CIFS protocol supports files that are created on unix, therefore windows can at least show a file with trailing dots created in Unix, it may not be able to operate on it though (I am not sure about that).

harendra-kumar:
@joehillen your comment prompted another question - what should we do to `dropExtension "foo."`? Should we drop the "." or not?

joehillen:
@harendra-kumar I think it should drop the dot.

That's consistent with how I've typically seen extensions removed in bash:
```bash
~$ file1=foo.txt
~$ echo "${file1%.*}"
foo
~$ file2=foo.tar.gz
~$ echo "${file2%.*}"
foo.tar
~$ file3=foo.
~$ echo "${file3%.*}"
foo.
```

harendra-kumar:
Sounds good to me. I cannot come up with any sound logical argument either way.

PR #114, opened 2018-01-22
Add stripFileExtensionFunction

sergv:
I noticed that `filepath` package has [stripExtension function](https://hackage.haskell.org/package/filepath-1.4.2/docs/System-FilePath-Posix.html#v:stripExtension), but `path` doesn't have anything like that. I have added it in this PR. I also took the liberty of unifying implementations of `addFileExtension`, 'removeFileExtension' and the new `stripFileExtension` functions. I have also submitted some tweaks to make package work with great `doctest` package and added tests for the `stripFileExtension` function.

### DIFF
--- a/src/Path/Include.hs
+++ b/src/Path/Include.hs
@@ -100,6 +100,10 @@ import           Language.Haskell.TH.Quote (QuasiQuoter(..))
 import           Path.Internal
 import qualified System.FilePath.PLATFORM_NAME as FilePath
 
+-- Setup code for doctest.
+-- $setup
+-- >>> :set -XTemplateHaskell
+
 --------------------------------------------------------------------------------
 -- Types
 
--- a/src/Path/Include.hs
+++ b/src/Path/Include.hs
@@ -343,7 +343,7 @@ fileExtension = FilePath.takeExtension . toFilePath
 -- | Add extension to given file path. Throws if the
 -- resulting filename does not parse.
 --
--- >>> addFileExtension "txt $(mkRelFile "foo")
+-- >>> addFileExtension "txt" $(mkRelFile "foo")
 -- "foo.txt"
 -- >>> addFileExtension "symbols" $(mkRelFile "Data.List")
 -- "Data.List.symbols"
--- a/src/Path/Include.hs
+++ b/src/Path/Include.hs
@@ -91,8 +91,10 @@ import           Control.Monad.Catch (MonadThrow(..))
 import           Data.Aeson (FromJSON (..))
 import qualified Data.Aeson.Types as Aeson
 import           Data.Data
+import           Data.Functor.Identity
 import           Data.List
 import           Data.Maybe
+import qualified Data.Traversable
 import           Language.Haskell.TH
 import           Language.Haskell.TH.Quote (QuasiQuoter(..))
 import           Path.Internal
@@ -361,12 +363,8 @@ addFileExtension :: MonadThrow m
   => String            -- ^ Extension to add
   -> Path b File       -- ^ Old file name
   -> m (Path b File)   -- ^ New file name with the desired extension added at the end
-addFileExtension ext (Path path) =
-  if FilePath.isAbsolute path
-    then liftM coercePath (parseAbsFile (FilePath.addExtension path ext))
-    else liftM coercePath (parseRelFile (FilePath.addExtension path ext))
-  where coercePath :: Path a b -> Path a' b'
-        coercePath (Path a) = Path a
+addFileExtension ext =
+  liftM runIdentity . liftFilePathFunc (Identity . (`FilePath.addExtension` ext))
 
 -- | A synonym for 'addFileExtension' in the form of an operator.
 -- See more examples there.
@@ -392,12 +390,8 @@ setFileExtension :: MonadThrow m
   => String            -- ^ Extension to set
   -> Path b File       -- ^ Old file name
   -> m (Path b File)   -- ^ New file name with the desired extension
-setFileExtension ext (Path path) =
-  if FilePath.isAbsolute path
-    then liftM coercePath (parseAbsFile (FilePath.replaceExtension path ext))
-    else liftM coercePath (parseRelFile (FilePath.replaceExtension path ext))
-  where coercePath :: Path a b -> Path a' b'
-        coercePath (Path a) = Path a
+setFileExtension ext =
+  liftM runIdentity . liftFilePathFunc (Identity . (`FilePath.replaceExtension` ext))
 
 -- | A synonym for 'setFileExtension' in the form of an operator.
 --
@@ -408,6 +402,19 @@ infixr 7 -<.>
   -> String            -- ^ Extension to set
   -> m (Path b File)   -- ^ New file name with the desired extension
 (-<.>) = flip setFileExtension
+
+liftFilePathFunc
+  :: (MonadThrow m, Data.Traversable.Traversable f)
+  => (FilePath -> f FilePath)
+  -> Path a File
+  -> m (f (Path a File))
+liftFilePathFunc f = \(Path path) ->
+  if FilePath.isAbsolute path
+  then Data.Traversable.mapM (liftM coercePath . parseAbsFile) $ f path
+  else Data.Traversable.mapM (liftM coercePath . parseRelFile) $ f path
+  where
+    coercePath :: Path x File -> Path x' File
+    coercePath (Path x) = Path x
 
 --------------------------------------------------------------------------------
 -- Parsers
--- a/src/Path/Include.hs
+++ b/src/Path/Include.hs
@@ -61,6 +61,7 @@ module Path.PLATFORM_NAME
   ,(<.>)
   ,setFileExtension
   ,(-<.>)
+  ,stripFileExtension
    -- * Parsing
   ,parseAbsDir
   ,parseRelDir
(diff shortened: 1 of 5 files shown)


NorfairKing:
@harendra-kumar I would like to hear your opinion on this.

I'm okay with this, but I want to see these tests:

- [ ] `stripFileExtension ".vimrc" ".vimrc"`
- [ ] `stripFileExtension ".foo" "..foo"`
- [ ] Property tests that the result is always valid

joehillen:
@sergv There is an open PR for this https://github.com/commercialhaskell/path/pull/113

sergv:
@joehillen I agree, the #113 could provide the missing bits to implement `stripFileExtension`. Still, I'd like to have that function in the API to match `filepath` package.

src/Path/Include.hs
```diff
@@ -409,6 +408,61 @@ infixr 7 -<.>
...
+stripFileExtension
+  :: MonadThrow m
+  => String                  -- ^ Extension to strip
+  -> Path b File             -- ^ Old file name
+  -> m (Maybe (Path b File)) -- ^ New file name without extension
+stripFileExtension ext =
+#if MIN_VERSION_filepath(1,4,1)
+  liftFilePathFunc (FilePath.stripExtension ext)
+#else
+  liftFilePathFunc stripExt
```
harendra-kumar: There is risk of these two implementations getting out of sync if the filepath implementation changes. To avoid that we can perhaps just remove the CPP conditional and keep the full implementation here.

src/Path/Include.hs
```diff
@@ -409,6 +408,61 @@ infixr 7 -<.>
...
+-- Just "foo"
+-- >>> stripFileExtension "foo.bar.baz" $(mkRelFile ".foo.bar.baz")
+-- *** Exception: InvalidRelFile ""
+--
+-- @since 0.6.2
+stripFileExtension
+  :: MonadThrow m
+  => String                  -- ^ Extension to strip
+  -> Path b File             -- ^ Old file name
+  -> m (Maybe (Path b File)) -- ^ New file name without extension
```
harendra-kumar: This should be just `m (Path b File)`. The `Maybe` is unnecessary because we have `MonadThrow m`.

harendra-kumar:
The more I think about it the more I dislike the idea of treating extensions as a concept/abstraction the way we are doing it. By having this concept we are treating the "." in a filename in a special way even though there is nothing special about it. I do not like the idea of treating ".vimrc" in a special way, for example, is it a file with extension or a file with no extension? What about filenames like "a." and "a..b"? It becomes ugly to handle cases like these.

Looking at the filepath package implementation of `stripExtension` - it says - "Drop the given extension from a FilePath, and the "." preceding it". Except in this case:

```
stripExtension ".c.d" "a.b..c.d" == Just "a.b."
``` 

It does not drop the preceding `.`. Also I do not like the fact whether you use ".ext" or "ext" the result is the same.

This all seems like a mess to me. What is the right solution then? Why not just drop all the APIs working one extensions and just use the following three APIs:

```
<.>
stripProperSuffix
isProperSuffixOf
```

One way to convince ourselves that these are in fact better APIs for operating on file extensions is that we have their exact counterparts operating on dir prefixes i.e. `</>`, `stripProperPrefix`, `isProperPrefixOf`.  It is befitting that we have duals of these for operating on file extensions. When we have duals we know that the whole picture fits in properly. 

However, I know that the duality is not perfect as we do not have a strict file extension counterpart of  the dir separator`/`. And "." is not that counterpart because ".." does not behave the same way in filenames as "//" does for directories paths.

Are there any downsides/pitfalls if we do something like this?

harendra-kumar:
We will still need `getExtension` and `dropExtension` functions though, and we again have the issue  of how to treat `.foo` and `foo.` :-(.

src/Path/Include.hs
sergv: True, but then if the `filepath` package moves on then `path`'s behaviour would diverge, which doesn't seem right to me.

src/Path/Include.hs
sergv: The `filepath` package returns Nothing if extension didn't match. It seems a bit too much to throw exception in such a common case.

sergv:
I have not used `path` package much yet, so I may have got this wrong. However, coming from `filepath` I'd expect at least some kind of compatibility with it, in particular within areas like what's considered an extension of paths like `.vimrc`, namely empty base name with ".vimrc" being the extension. It seems to me that if `path` is to adopt the approach you're proposing with not treating dot specially (which *does* seem elegant) it should clearly state somewhere in the documentation or README that there are differences from other libraries and that `path` is not a drop-in replacement for any other library.

src/Path/Include.hs
harendra-kumar: Whenever we use the `filepath` routines we are saying that the behavior of `path` depends on the version of `filepath` you are choosing. So the behavior of path may change even without a change of version, this is a bit disconcerting. This is a general problem though since we are using `filepath` routines at other places as well.

src/Path/Include.hs
harendra-kumar: The `Maybe` instance of `MonadThrow` takes care of that:

```
instance MonadThrow Maybe where
  throwM _ = Nothing
```

So we have:

```
Prelude Path Control.Monad.Catch> f = throwM (InvalidAbsDir "x") :: MonadThrow m => m Int
Prelude Path Control.Monad.Catch> f :: Maybe Int
Nothing
```

Take a look at `addFileExtension` it also throws an exception. We need exactly the same signature here as well.

harendra-kumar:
`path` is typed and `filepath` is untyped which means `path` cannot always do arbitrary things whereas `filepath` can. For example, in `path` an empty file name is invalid whereas it does not matter in `filepath` because filename is just a string. This means in case of `.vimrc` we cannot drop "the extension" because that means we have to return an empty filename which is invalid. So we have to be incompatible with filepath.

NorfairKing:
@sergv Paths are hard. See the parts about extensions here: https://cs-syd.eu/posts/2017-10-28-the-madness-of-paths

`path`'s API's semantics are entirely defined by what is tested to be the case, not via the semantics of `filepath` (because they are sometimes wrong or incompatible with strong typing).

src/Path/Include.hs
sergv: Sorry, I wasn't clear enough. If Maybe will be removed then an exception in `MonadThrow m => m` will be used so signal either that resulting path is invalid or that suffix didn't match. If both these use cases are merged into single `MonadThrow m` then user may choose `Maybe` for `m` and get Nothing when resulting path is invalid and doesn't parse or using `m` as `IO` or something similar he'll get an exception when the suffix was not present. I believe that if resulting path is invalid then a loud error/exception should be produced.

Perhaps function's semantics should be changed to always return either the original path or a path without a given extension.

harendra-kumar: We do not have two distinct cases here. Either we are able to strip the extension, resulting in a valid filepath and extension (Just case), or not (Nothing case). I cannot think of a case where we have to distinguish the error case in two different categories. Am I missing something?

harendra-kumar: Looking at the test cases, this is the case where you are throwing an exception:

```
stripFileExtension ".foo" $(mkRelFile ".foo")
```

There are two ways of looking at this. First is to treat ".foo" as an extension to an invalid filename i.e. an empty filename. The second is to treat ".foo" as a filename without an extension rather than an extension, the "." is part of the filename in this case rather than being an extension separator. You have chosen the former way. I think we can choose the latter and keep the API simple. Two layers of error handling makes the API unnecessarily complex.

harendra-kumar:
Extensions are loosely defined based on conventions and those conventions are less strict rules compared to the conventions for the path separator (`/`).  I tried to formulate some rules based on a more or less consistent reasoning. Here are three simple rules of reasoning:

1) "." and "" are not valid as filenames. This is not a new rule but a fact already observed by the `path` package.
2) A filename always consists of a *mandatory* and *valid* filename followed by an *optional* *valid extension*. This is just a formal way of saying what we already know in practice.
3) What is a valid extension?
a) A valid extension is also a valid filename. A corollary of this is that "."  and "" are not valid extensions just as they are not valid filenames. This is also nothing new.
b) a valid extension always starts with single "." e.g. `..foo` is not a valid extension but `.foo` is valid. This formalizes the resolution of an ambiguity about where the dots go when there are multiple dots in the path name. This rule says that when ambiguous they are treated as part of the filename and not the extension. Again this is pretty intuitive and nothing new.

All these rules are not new but if we follow them consistently we get a consistent scheme as I outlined in the table below. You can verify that the whole table follows these simple rules. 

| Combined      | Filename   | Extension |
|-------------- | -----------|-----------|
| `.foo`             | `.foo`        |                  |
| `..foo`            | `..foo`       |                  |  
| `foo.`             | `foo.`        |                  | 
| `foo..`            | `foo..`       |                  |  
| `foo.bar`        | `foo`         |  `.bar`            |  
| `foo.bar.`       | `foo`         | `.bar.`            |
| `foo.bar..`      | `foo`         | `.bar..`           |
| `foo..bar`       | `foo.`        | `.bar`             |
| `foo...bar`      | `foo..`       | `.bar`             |
| `foo.bar.qux` | `foo.bar`  | `.qux`             |
| `foo.bar.qux` | `foo`         | `.bar.qux`      |

Some explanations:

* `.foo` is a filename without an extension because as per rule 2 a filename must have a valid filename part followed by an optional extension.
* `..foo` is a filename because if we treat `.foo` as an extension we are left with an invalid filename `.` but as per rule 2 we need a valid mandatory filename.
* `foo.` is a filename because "." or "" are not valid extensions.

harendra-kumar:
In most cases `filepath` agrees with this table with the following two exceptions:

1) Leading dot case  e.g. `.foo`, filepath treats the file as an extension and returns an empty filename.
2) `filepath` treats multiple leading dots as a valid extension e.g. `..bar` is a valid extension. This makes the `foo..bar` case ambiguous i.e. it can be `foo.` + `.bar` as well as `foo` + `..bar`. However, `dropExtension` needs to make a choice and drops `.bar`.

harendra-kumar:
To move forward on this PR, tldr of my very long comments above is:
* have those two cases  (`.foo` and `..foo`) handled correctly and have tests for those.
* it will be nice to remove ambiguity in the other cases I mentioned but we can live with that.

With this and if we resolve this one comment https://github.com/commercialhaskell/path/pull/114#pullrequestreview-90931087, it should be okay to merge this PR in my opinion, once the maintainers clear it.

sergv:
Okay, at the very least I'll cover `.foo` and `..foo` before submitting an update. I feel like I should play a bit trying to remove ambiguity in other cases and see where that leaves the code and how bad would it get. I'll to that in the near future and come back to you.

harendra-kumar:
It can be removed if we follow rule 3b in my comments above i.e. by treating `..foo` to be invalid as an extension. In case of `stripFileExtension` it means we should not strip "..foo" . On the other hand, in case of `addFileExtension` it means we should not add "..foo" as an extension to make sure that `stripFileExtension . addFileExtension` is always `id`.  

However, I am not sure if it is worthwhile to remove the ambiguity. Since extensions are loosely defined, in general people may have the expectation that adding or removing "..foo" should be allowed. The only thing we gain with this is making `dropFileExtension . addFileExtension` idempotent.

As a side note, I think it is a good idea to also have tests to ensure that `stripFileExtension . addFileExtension` is `id`.

sergv:
The reasoning outlined so far is all well and good, but I just tried out running `fileExtension` function and it considers extension of `.vimrc` to be `.vimrc`.

```
> pure . fileExtension =<< parseRelFile ".vimrc"
".vimrc"
```

Thus `path` package does not adhere to the conventions outlined so far. I am happy to consider revisiting this PR once the relevant conventions will be implemented in the library. Otherwise, current alternatives of moving this PR forward do not seem right. On the one hand, incorporating all the comments will make `stripFileExtension` function behave differently than the rest of the API, which I would like to avoid. On the other hand, making all functions treat extensions as outlined before will involve a major rewrite of this package, which is way out of scope for this PR. 

A simpler way to implement `stripFileExtension` should go through `fileExtension` and `fileBasename`  (or equivalent) functions. It would be consistent with how they treat extensios and not depend on `filepath` version. I can make the relevant change once `fileBasename` will be added to the API. Until then it appears there's no way I can make further progress on this PR. Feel free to close and I'll open a new one once API starts providing the neccessary function.

harendra-kumar:
I just sent PR #115 that makes the existing extension operations strictly lawful. As part of that I added `splitExtension` operation which I did not expose deliberately. We can expose it later once we sink in the basic changes in that PR. With `splitExtension` we should not need `fileBaseName` or `stripFileExtension`. Both of these operations can be trivially derived from `splitExtension`.

sergv:
I think this can be closed in favour of #115.

PR #119, opened 2018-03-21
More strict checking on CI

mrkkrp:


NorfairKing:
I'm not convinced that it's sound to turn off optimisation for `dev`. What do you think?

mrkkrp:
Why it is not sound?

mrkkrp:
The reason it is turned off is that it makes builds faster, which is nice for CI.

mrkkrp:
Why would optimization level influence anything that matters on CI? Only logic itself matters for tests, and other checks like `-Wall -Werror -Wcompat` have nothing to do with optimizations.

joehillen:
I think NorfairKing is concerned about the fact that GHC's optimizer has had bugs in the past, but I think that's outside the scope of this library.

NorfairKing:
@joehillen Correct

@mrkkrp I'm happy with this PR then. +1 from me.

mrkkrp:
Let's have some faith in GHC, shall we? Bugs can be everywhere, after all :)

PR #122, opened 2018-04-27
Expose splitExtension operation

harendra-kumar:
1) Breaking - Change the return type of `fileExtension` to throw an exception when there is no extension

2) Now all the extension operations can be expressed elegantly in terms of simple laws, all of them verified by property tests. `addExtension` and `splitExtension` are the basic operations and the rest are just convenient and trivial applications of these two. `addExtension` and `splitExtension` are lawful inverses of each other with this change.

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -1,7 +1,10 @@
 UNRELEASED:
+	* BREAKING CHANGE: "fileExtension" now throws an exception if the file has no
+		extension. You can use the result as a "Maybe" in pure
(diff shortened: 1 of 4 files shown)


harendra-kumar:
I have added Since 0.7.0 annotations for the new operations anticipating 0.7.0 to be the next release.

harendra-kumar:
@NorfairKing @mrkkrp

NorfairKing:
Love it :+1:

PR #140, opened 2019-01-28
A sum type uniting absolute and base paths, for parsing

mstksg:
Thank you for the great library! :)

I've had some situations where I've had to parse a filepath as "either absolute or relative", and have had to branch on one situation or the other.  I've pulled together some of the common logic as a type in case you might be interested.  The main type is:

```haskell
data SomeBase t = Abs (Path Abs t)
                | Rel (Path Rel t)
```

which is mostly useful for parsing:

```haskell
parseSomeDir :: MonadException m => FilePath -> m (SomeBase Dir)
parseSomeFile :: MonadException m => FilePath -> m (SomeBase File)
```

Users can parse a `FilePath` representing a `Dir` or a `File`, and then branch on whether the path was abs or relative.

If this doesn't really fit into the library as it is, feel free to close :)


NorfairKing:
I've needed this in the past, so I wouldn't mind.
I'm not much fussed to bikeshed the name, but I'd definitely like to hear the opinion of the other contributors.

chrisdone:
I like this, +1 from me.

NorfairKing:
@mstksg LGTM then. We need a rebase though.

mstksg:
Rebase completed :)  Note that the stack resolver in the stack.yaml won't compile as-is, since it is tied to a GHC older than the base bounds.

NorfairKing:
@mstksg  then I guess you're blocked on #147

jecaro:
Just wanted to let you know that I'd love to see PR merged into master. Also thanks for such an awesome library.

chrisdone:
The merge conflict seemed trivial so I resolved it. The compile might still fail, but that's fine. This PR seems to have stalled but not for any reason, so waking it up again. Thanks @jecaro

chrisdone:
Tests passed, merged! Thanks @mstksg. Sorry about the year-long delay.

mstksg:
no worries, happy i was able to contribute :D

PR #146, opened 2019-08-28
Made a working stack build again

NorfairKing:


mrkkrp:
The build with stack fails, but shouldn't. It appears that `genvalidity-0.3.2.0` should allow newer `QuickCheck`. Are you going to fix it somehow?

NorfairKing:
Oh you're right, it does. This is not worth fixing seperately then. Let's just go with the `upgrade-deps` branch.

PR #157, opened 2020-05-07
Export SomeBase data constructor

jecaro:
Following #140 PR, it misses the export of the data constructor. Otherwise there is not much we can do with `SomeBase`.


NorfairKing:
LGTM

PR #160, opened 2020-05-15
Fixes #159

### LINKED ISSUE #159: QuasiQuoters and TH smart constructors are broken
Since the addition of `Lift` instances (commit [68314ea70b670f156975049505015ac94dd66557](https://github.com/commercialhaskell/path/commit/68314ea70b670f156975049505015ac94dd66557)) the following is possible:

    foo :: Path Abs Dir
    foo = [relfile|foo|]

Since the quasi quotes use the `mk{Abs,Rel}{Dir,File}` functions under the hood it applies to them, too.

This is IMHO clearly a bug.

mmhat:
Provided a path of type `Path a b` we use the `Typeable` instances of
`a` and `b` to insert the correct type in the splice.

### DIFF
--- a/test/TH.hs
+++ b/test/TH.hs
@@ -1,0 +1,58 @@
+{-# LANGUAGE MultiParamTypeClasses #-}
+{-# LANGUAGE QuasiQuotes #-}
+{-# LANGUAGE TemplateHaskell #-}
+
+module TH where
+
+import qualified Language.Haskell.TH.Syntax as TH
+import Path
+import Path.Internal
+
+
+
+class Foo a b where
+    foo :: Path a b -> FilePath
+    foo = toFilePath
+
+instance Foo Abs Dir
+instance Foo Abs File
+instance Foo Rel Dir
+instance Foo Rel File
+
+
+
+qqAbsDir :: FilePath
+qqAbsDir = foo [absdir|/foo/|]
+
+qqAbsFile :: FilePath
+qqAbsFile = foo [absfile|/foo|]
+
+qqRelDir :: FilePath
+qqRelDir = foo [reldir|foo/|]
+
+qqRelFile :: FilePath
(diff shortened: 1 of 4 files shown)


test/TH.hs
```diff
@@ -0,0 +1,58 @@
...
+liftAbsDir = foo $(TH.lift (Path "/foo/" :: Path Abs Dir))
+
+liftAbsFile :: FilePath
+liftAbsFile = foo $(TH.lift (Path "/foo" :: Path Abs File))
+
+liftRelDir :: FilePath
+liftRelDir = foo $(TH.lift (Path "foo/" :: Path Rel Dir))
+
+liftRelFile :: FilePath
+liftRelFile = foo $(TH.lift (Path "foo" :: Path Rel File))
```
NorfairKing, line 58: I'm very happy to see that you added more tests.
We may want to add a comment here that this compile-time test works the way you want it because if you had the bug then this would give ambiguities.
I guess there's no way to test that a certain thing _doesn't_ compile.

mmhat: Ok, I'll put some comments in there.
With path-0.7.0 (where the bug is present) it indeed doesn't compile:
Ambiguous type variables ‘a0’, ‘b0’ arising from a use of ‘foo’

I.e. GHC does not know which instance of `Foo` to use.

NorfairKing: :+1: This seems like a clear improvement to me then.

mmhat: Ok, I'm fine with that.

mmhat:
@NorfairKing I'd be happy if I see a new minor release rather sooner than later due to the severity of the bug. Could you do this?

NorfairKing:
@mmhat I can't, but @chrisdone can do it I think.

chrisdone:
I pushed a new release 0.7.1.

PR #162, opened 2020-05-22
Add MonadError in addition to MonadThrow.

### LINKED ISSUE #149: MonadThrow vs MonadError
The problem with using `MonadThrow` is that the caller can theoretically expect just about any exceptions to be thrown. Whereas with `MonadError e`, the now-small range of errors that the caller can expect to handle is known at the type level (and, not to mention, in the function documentation). 

Would you be open to a PR changing all the functions that currently use `MonadThrow m` to using `MonadError PathError m`?

sellout:
This abstracts Include.hs to be able to swap between `MonadError` and
`MonadThrow`. The `MonadError` implementations are exposed via
`Path.Except[.*]`, while `MonadThrow` remains in `Path[.*]`.

Some shortcomings and considerations:
- there is now a dependency on MTL
- `-Wunused-imports` is disabled so both classes can be imported
  without having to conditionalize anything
- there are no tests run against `Path.Except[.*]` yet

A slightly different approach could be to publish two packages -- `path`
and `path-except`. They would expose the exact same modules (`.Except`
would disappear), but one would have a dependency on `exceptions` and
the other on `mtl`. It's unlikely that anyone would want to use both the
MonadThrow and MonadError approaches in the same project, so this could
reduce their dependency footprint, but it does mean I have to fix the
unused imports.

Fixes #149.

### DIFF
--- a/path.cabal
+++ b/path.cabal
@@ -23,6 +23,9 @@ flag dev
 library
   hs-source-dirs:    src
   exposed-modules:   Path
+                   , Path.Except
+                   , Path.Except.Posix
+                   , Path.Except.Windows
                    , Path.Internal
                    , Path.Posix
                    , Path.Windows
@@ -32,6 +35,7 @@ library
                    , exceptions >= 0.4     && < 0.11
                    , filepath   < 1.2.0.1  || >= 1.3
                    , hashable   >= 1.2     && < 1.4
+                   , mtl
                    , text
                    , template-haskell
   if flag(dev)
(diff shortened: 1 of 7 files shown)


NorfairKing:
* The lacking tests are a blocker for me.
* I don't care about `-Wunused-imports` as long as there is no `-Werror` and `-Wall` is still on.
* I would also be ok with a separate package. What do other maintainers think of this?

sellout:
I think I would _prefer_ the separate packages, personally. And yes, I will definitely get the tests working before this is merged, but I won’t get a chance until Tuesday and didn’t want to hold up any discussion.

NorfairKing:
I think separate packages would also be easier for us. I'm just not sure how you would organise the code in that case, but I'll let you figure that out.

locallycompact:
Did the new package happen? I'd quite like to try this.

PR #164, opened 2020-07-20
Add changeDir function

locallycompact:
Hi, I use this a lot and it's pretty simple so I wondered if it is in scope of this library.


NorfairKing:
@locallycompact I would probably call it `replaceProperPrefix`,  would that be acceptable?

locallycompact:
Makes sense to me.

PR #168, opened 2020-10-18
Fixed appveyor.yml

Martinsos:
This is draft for now because I need a PR to test the CI. Once I get it running I will convert it into proper PR.


Martinsos:
Appveyor build is still failing, but I believe this is now due to the actual bug in Path crashing it. Since bugs should be fixed in my other PR, I will close this PR and move this fix there, where I have the change of getting it to pass.

PR #179, opened 2021-07-08
Added strong-path to list of similar libraries.

Martinsos:
Since strong-path is very similar to path and is tightly connected to it, I thought it might be a good addition to the list of similar libraries!

### DIFF
--- a/README.md
+++ b/README.md
@@ -524,6 +524,20 @@ and there are no functions like parent. But there’s not much to complain
 about. It’s a fine library, but I didn’t feel the need to drop my own in
 favor of it. Check it out and decide for yourself.
 
+### strong-path
+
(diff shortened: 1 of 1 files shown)


NorfairKing:
@Martinsos I'm not opposed to this, but this section:
```
## Comparing with existing path libraries
```
was about why this library needs to exist.

Martinsos:
@NorfairKing ah, that makes sense! I saw it more as a comparison, but now that I look again I see what you mean, it is explanation of what other libraries were missing that `path` did better. Except for the last one maybe which was discovered a bit later hm.

No problem, I thought this would be a good opportunity to spread the word about strong-path as a possible alternative solution (if even more type safety is needed), but I don't want to crowd the docs, especially if there is no appropriate section!

PR #181, opened 2021-11-15
Add mapSomeBase and prjSomeBase for modifying or projecting SomeBase.

kquick:
These are helper functions that allow the modification of a SomeBase
value (e.g. `parent`) or projecting the value out of the SomeBase via
a projection function (e.g. `toFilePath`).

### DIFF
--- a/src/Path/Include.hs
+++ b/src/Path/Include.hs
@@ -18,12 +18,14 @@
 -- we represent the notion of a relative root by "@.@". The relative root denotes
 -- the directory which contains the first component of a relative path.
 
-{-# LANGUAGE TemplateHaskell #-}
-{-# LANGUAGE PatternGuards #-}
 {-# LANGUAGE DeriveDataTypeable #-}
 {-# LANGUAGE DeriveGeneric #-}
 {-# LANGUAGE EmptyDataDecls #-}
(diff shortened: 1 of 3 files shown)


NorfairKing:
Looks good. Let's cover these with tests as well.

kquick:
I had planned to... but there don't seem to be _any_ tests for `SomeBase`.

kquick:
I can add some tests that involve SomeBase, but I would probably take the approach to adding them to the `ValidityTest.hs` which would involve adding generators to `test/Path/Gen.hs`.  I had  held off because that was a much bigger impact to the change, but if that's desireable and a reasonable approach I could add those.

NorfairKing:
> I can add some tests that involve SomeBase, but I would probably take the approach to adding them to the ValidityTest.hs which would involve adding generators to test/Path/Gen.hs. I had held off because that was a much bigger impact to the change, but if that's desireable and a reasonable approach I could add those.

Yes we should do this. I wish I had more time to get involved properly and help with this.

NorfairKing:
@kquick Thanks! I'll take it from here! Please give me some time to get to this.

NorfairKing:
Merged in #185

Issue #186, opened 2021-12-24
Adding 'Data' instance to Rel, File, Abs and Dir


Martinsos:
Is there are a particular reason why Rel, File, Abs and Dir are not deriving `Data`?

`data Path b t` is deriving Data, and what that does is create an instance Data for it where `b` and `t` also need to have instances of Data. However, since Rel, File, Abs and Dir don't have `Data` instances, this means that `Path Rel File`, `Path Rel Dir`, ... and so are not instances of `Data`.

I am encountering this is a problem now in https://github.com/wasp-lang/strong-path because I would like to make main type of StrongPath instance of `Data`, but the only thing stopping me from deriving `Data` is the fact that internally StrongPath uses `Path Rel File`, `Path Rel Dir` and so on, and those are not instances of `Data`.

I believe making `Rel`, `File` and similar instances of `Data` is as simple as just adding `Data` to `deriving`.

NorfairKing:
Sounds agreeable!

Martinsos:
> Sounds agreeable!

Ok! I created a very simple PR here https://github.com/commercialhaskell/path/pull/187 -> I think this should be it and I can see all the checks passing. Would love to get your eyes onto it also, just to make sure I haven't missed something.

NorfairKing:
This is long-fixed. Just closing this issue now.

NorfairKing:
This is long-fixed. Just closing this issue now.

PR #187, opened 2021-12-24
Make Rel, Abs, File and Dir instances of Data

Martinsos:
As I describe in https://github.com/commercialhaskell/path/issues/186, this would enable `Path Rel Dir`, `Path Rel File`, and so on, to also be instances of `Data`, which enables any other data type that uses `Path` as part of its structure to also easily derive `Data`.


NorfairKing:
Nothing disagreeable from my perspective. I'll merge in a few days if no one else objects.

PR #191, opened 2023-10-16
Fix #190 Add splitDrive, takeDrive, dropDrive and isDrive

### LINKED ISSUE #190: Add `takeDrive :: Path Abs Dir -> Path Abs Dir`
The `filepath` package provides `System.FilePath.takeDrive :: FilePath -> FilePath`. It would be useful to have a well-typed version.

mpilgrem:
Also adds related tests in test-suite `test`.

Also adds related tests in test-suite `validity-test`, but `validity-test` (generally, before this change) hangs when run on my Windows 11 machine (so I could not test the tests), and I note that `validity-test` is excluded from the existing CI.

Also adds an entry in the Change Log and bumps the package version.

### DIFF
--- a/src/Path/Include.hs
+++ b/src/Path/Include.hs
@@ -63,6 +63,10 @@ module Path.PLATFORM_NAME
   ,splitExtension
   ,fileExtension
   ,replaceExtension
+  ,splitDrive
+  ,takeDrive
+  ,dropDrive
+  ,isDrive
   ,mapSomeBase
   ,prjSomeBase
    -- * Parsing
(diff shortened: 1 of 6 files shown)


mpilgrem:
The CI is failing because it is out of date. I will add a commit to update the CI to this pull request.

src/Path/Include.hs
```diff
@@ -377,6 +381,31 @@ parent (Path fp) =
     $ FilePath.takeDirectory
     $ FilePath.dropTrailingPathSeparator fp
 
+-- | Split an absolute path into a drive and, perhaps, a path. On POSIX, @/@ is
+-- a drive.
```
NorfairKing, line 385: >  On POSIX, @/@ is a drive.

Is it a good idea to have these functions defined for non-windows?
We could put it in the windows-specific code I guess, but I don't have an opinion yet about whether that's a good idea.

NorfairKing (approved):
This LGTM, I only left one question about whether we want these functions to be available on POSIX.

src/Path/Include.hs
mpilgrem: Non-well typed equivalent functions exist for both POSIX and WINDOWS at `filepath` package: [`System.FilePath`](https://hackage.haskell.org/package/filepath-1.4.100.4/docs/System-FilePath.html#g:5). I think it is sensible to prove well-typed equivalents. I followed that package's Haddock documentation for this package.

NorfairKing:
@mpilgrem Ok, I'll merge this.

Issue #136, opened 2018-10-18
aeson dependency


2mol:
Hi everybody! First of all, this library is really great, and I very much enjoy using it.

I wanted to discuss the dependency on aeson. As far as I can see, the only weight it's pulling, is a `toJSON` instance in `Path.Internal`. Since aeson itself has a pretty massive dependency graph, that's a lot of dependency for not a whole lot of feature. I mean, aeson is super nice, but I don't use it in every program.

Let me know if there was already a discussion about this, or if there are some obvious good reasons like avoiding orphan instances that counterbalance my desire for faster initial compilation times. The last thing I want to do is nitpick something good. But personally, I would be really happy if my dependency graph could be trimmed of aeson it's dependencies.

NorfairKing:
@2mol There are a few options:

- The current situation: `path` with an `aeson` dependency
- Make `aeson` depend on `path` and give it a non-orphan instance. (I would be happy with this, but not sure if you'll be able to convince the `aeson` maintainers. You'll have to open an issue there and refer back if you want to try this.)
- Make two seperate packages: `path` and `path-aeson` where the latter has the instances which would be orphan instances.
- Make the `aeson` dependency flag-activated so you can turn it off for your project. This would mess with a bunch of tooling if it's off by default, but we can leave it on by default too.

It would be worth discussion these points

2mol:
Yeah, I looked at how many transitive dependencies `aeson` would gain by depending on `path`, and the direct ones are just two: `filepath` and `exceptions`. Zero would have been nicer, but I asked people if they would be open to take that tradeoff: https://github.com/bos/aeson/issues/677

chrisdone:
If it helps to add some direction to this discussion, I think we should avoid conditional compilation where possible. A package is easiest to understand, test and tool when it is just a package, and not N packages depending on configuration.

I can understand not wanting to depend on `aeson` for some smaller tools. So that leaves only a `path-aeson` instances package.

**How this affects library users:** they need to add `path-aeson` to their .cabal file. That's a pretty easy step.

However, how do we inform end-users when they eventually see `No instance for FromJSON (Path ..)` that this error is fine and to simply add this dependency? Bite the bullet?

chshersh:
[...]

PR #192, opened 2024-06-21
Support `AbstractFilePath`

### LINKED ISSUE #189: Support `AbstractFilePath`
## Motivation

The next filepath release will add support for a new API, which fixes subtle encoding issues and improves cross platform code and memory residence.

You can read about the state of the newly added API here: https://github.com/haskellfoundation/tech-proposals/issues/35

Release candidate with haddock is here: https://hackage.haskell.org/package/filepath-2.0.0.3/candidate

Demonstration about unsoundness of base with exotic encodings and how the new API fixes them is here: https://gist.github.com/hasufell/c600d318bdbe010a7841cc351c835f92

## Migrations

Migrations can happen once the `filepath`, `unix` and `Win32` packages are updated and in sync. A migration would usually involve using the new types from [System.AbstractFilePath](https://hackage.haskell.org/package/filepath-2.0.0.3/candidate/docs/System-AbstractFilePath.html) and the new API variants from [unix](https://github.com/haskell/unix/pull/202)/[Win32](https://github.com/haskell/win32/pull/198).

When writing low-level cross platform code manually (shouldn't generally be necessary), the usual strategy is this:

```hs
#if defined(mingw32_HOST_OS) || defined(__MINGW32__)
import qualified System.AbstractFilePath.Data.ByteString.Short.Word16 as SBS
import qualified System.AbstractFilePath.Windows as PFP
#else
import qualified System.AbstractFilePath.Data.ByteString.Short as SBS
import qualified System.AbstractFilePath.Posix as PFP
#endif

crossPlatformFunction :: AbstractFilePath -> IO ()
#if defined(mingw32_HOST_OS) || defined(__MINGW32__)
crossPlatformFunction (OsString pfp@(WS ba)) = do
    -- use filepath functions for windows specific operating system strings
    let ext = PFP.takeExtension pfp
    -- operate directly on the underlying bytestring (which is a wide character bytestring, so uses Word16)
    let foo = SBS.takeWhile
    ...
#else
crossPlatformFunction (OsString pfp@(PS ba)) = do
    -- use filepath functions for posix specific operating system strings
    let ext = PFP.takeExtension pfp
    -- operate directly on the underlying bytestring (which is just Word8 bytestring)
    let foo = SBS.takeWhile
    ...
#endif
```

Platform specific code can be written using [`PosixFilePath`](https://hackage.haskell.org/package/filepath-2.0.0.3/candidate/docs/System-AbstractFilePath-Posix.html)/[`WindowsFilePath`](https://hackage.haskell.org/package/filepath-2.0.0.3/candidate/docs/System-AbstractFilePath-Windows.html) types.

If you have further questions, please let me know. I'm going to write a blog post outlining the affairs and more in-depth intro and migration strategies close after the release. This is a heads up.

mmhat:
This PR adds a new namespace `OsPath` to the library which contains a version of `Path` that is a newtype wrapper around the `System.OsPath` filepaths. As a consequence all functions that took a `FilePath` as an argument or produced on, or that took/produced a `String` (e.g. extensions) now use the counterparts from `System.OsPath` and `System.OsString` respectively.
`toFilePath` is an exception to this, it still gives you a `System.FilePath.FilePath`.
For the implementation I aimed to stay as close to the existing `Path` modules as possible, that's why there are for example also ports of the deprecated functions included.

As it stands this PR consists of two parts: The first commits are cleanup/refactoring commits of the code that already exists.
Those happened either because something was missing (`hie.yaml`) or something got in my way (`.ghci.conf`) or - the biggest part - as a preliminary work before the replication of the various modules. For example, I refactored the `test` testsuite such that the copy could be easier adapted to the new `OsPath` hierarchy.

There are some ugly parts where I am not entirely sure if I found the right solution - in particular the `aeson` instances and everything involving conversion from one encoding to another, where I had to use `unsafeDupablePerformIO` at some places.

Gently pinging @hasufell @mpilgrem (I noticed to late that you started working on this issue too) and @NorfairKing since I am interested in your feedback specifically.

Fixes #189


NorfairKing:
@mmhat I'll gladly review this and probably get it merged eventually, but I need some time to review this carefully.

src/OsPath/Internal/Include.hs
```diff
@@ -0,0 +1,321 @@
...
+#elif MIN_VERSION_template_haskell(2,16,0)
+  liftTyped = TH.unsafeTExpCoerce . TH.lift
+#endif
+
+-- | Convert to a 'FilePath' type.
+--
+-- All directories have a trailing slash, so if you want no trailing
+-- slash, you can use 'System.FilePath.dropTrailingPathSeparator' from
+-- the filepath package.
+toFilePath :: Path b t -> FilePath
```
hasufell, line 151: IMO, you shouldn't provide this function. There are 3 functions in filepath/or-string to convert from `OsPath` to `FilePath`:

* decodeUtf
* decodeWith
* decodeFS

The decision here to use decodeFS ad-hoc without forcing the user to think about the choice subverts the point of the new API, I think.

Let them get a proper OsPath back and then have them deal with OsPath API.

NorfairKing:
@mmhat I just tried building and testing this locally.

Could you get `stack test --pedantic` to pass?

mmhat:
> Wrt aeson, also see:
> 
> * https://hasufell.github.io/posts/2024-05-07-ultimate-string-guide.html#to-json

I this blog post a while ago, but I really forgot that JSON marshalling is covered there!
So thanks for the pointer, and thanks for writing this in the first place.

> 
> Generally, we shouldn't need to ever assume an encoding. Using `decodeFS`/`encodeFS` can be a sign of a bug, but not always.

Ah, yes, you are right; I changed the JSON instances now such they assume Unicode encoding for the underlying filepath.

> Could you get `stack test --pedantic` to pass?

Done.

src/OsPath/Internal/Include.hs
mmhat: True, I removed the `toFilePath` function.

src/OsPath/Internal/Include.hs
```diff
@@ -0,0 +1,335 @@
...
+  {-# INLINE show #-}
+
+instance NFData (Path b t) where
+  rnf (Path x) = rnf x
+  {-# INLINE rnf #-}
+
+instance ToJSON (Path b t) where
+  toJSON =
+      either (error . displayException) toJSON
+    . OsPath.decodeUtf
```
hasufell, line 115: We can avoid the `error`, but it's up to debate:

* try `decodeUtf` (equals utf8 on unix and utf16le on windows)
* in case of failure, fall back to `latin1` on unix and `ucs-2` on windows, which both cannot fail (morally... the types of `decodeWith` lie in this case)

This is what I dislike about aeson. You can't really specify different encodings for the same type, unlike waargonaut.

mmhat: Sounds fine to me.
@hasufell Does your comment also apply to the encoding functions? Also, I found `System.OsString.Encoding.ucs2le`, but no `TextEncoding` for `latin1`.

hasufell: https://hackage.haskell.org/package/base-4.20.0.1/docs/System-IO.html#v:latin1

`encodeWith` would error on Chars outside of the range. There is very few, maybe no Encoding that is total under `encodeWith`, since you can represent Surrogates as Chars.

src/Path/Include.hs
```diff
@@ -482,9 +482,11 @@ splitExtension (Path fpath) =
             trailingSeps = takeWhile isSep rstr
             xtn  = (takeWhile notSep . dropWhile isSep) rstr
         in (reverse name, reverse xtn ++ trailingSeps)
-    normalizeDrive
-        | IS_WINDOWS = normalizeTrailingSeps
-        | otherwise  = id
+#if IS_WINDOWS
```
NorfairKing, line 485: What was this change for?

src/Path/Include.hs
```diff
@@ -833,22 +846,15 @@ normalizeWindowsSeps :: FilePath -> FilePath
...
-normalizeTrailingSeps :: FilePath -> FilePath
-normalizeTrailingSeps = reverse . normalizeLeadingSeps . reverse
+#endif
 
 -- | Applies platform-specific sep normalization following @FilePath.normalise@.
 normalizeFilePath :: FilePath -> FilePath
-normalizeFilePath
-  | IS_WINDOWS = normalizeWindowsSeps . FilePath.normalise
-  | otherwise  = normalizeLeadingSeps . FilePath.normalise
+#if IS_WINDOWS
```
NorfairKing, line 853: Same question here: Why?
If you're going to do this, it makes sense to put the type signature inside the if statement so that it's easier for formatters to figure out.

validity-test-ospath/OsPath/Gen/Include.hs
```diff
@@ -0,0 +1,145 @@
...
+-- be valid.
+instance GenValid PLATFORM_PATH where
+    -- We also need to exclude UTF-16 surrogates.
+    genValid = mconcat <$> listOf (OsString.unsafeEncodeUtf . (:[]) . chr <$> frequency
+        [ (2, choose (0x0, 0xD800 - 1))
+        , (2, choose (0xDFFF + 1, 0x10FFFF))
+        , (1, elements (map ord "./\\"))
+        ]
+        )
+    shrinkValid _ = [] -- TODO: Not yet implemented
```
NorfairKing, line 142: Didn't we have shrinking before?
This will be important if/when we run into bugs but it's only a maintenance burden so maybe not so important for users.

validity-test-ospath/OsPath/Gen/Include.hs
```diff
@@ -0,0 +1,145 @@
...
+
+instance Validity PLATFORM_PATH where
+    validate = trivialValidation -- TODO: Not yet implemented
```
NorfairKing, line 145: This definitely looks incomplete. I'd be surprised if these were trivially valid.

NorfairKing:
@mmhat Great job on this work. It must have been quite a fiddly thing to get right, but you got `stack test --pedantic` to pass as-is on my machine the first time. :tada: 

I had a look through the changes and don't see anything disagreeable.
I left a few nits but nothing strictly blocking I think.

I wasn't confident so I will just ask you: Can you please double-check that the original `Path` API isn't changed and no tests have been removed?

src/OsPath/Internal/Include.hs
mmhat: > https://hackage.haskell.org/package/base-4.20.0.1/docs/System-IO.html#v:latin1

Thanks; I missed that `System.IO.TextEncoding` and `GHC.IO.Encoding.Types.TextEncoding` are the same. The latter is linked in the Haddocks in `System.OsString.Encoding`, e.g. [here](https://hackage.haskell.org/package/os-string-2.0.3/docs/System-OsString-Encoding.html#v:ucs2le).

Regarding the `FromJSON/ToJSON` instances: The more I think about it the less I am convince that they should be in this library in the first place - Or at least not brought in scope by `OsPath`. There is clearly more than one sensible instance and throughout the ecosystem that is dealt with by leaving the instance definition to the user of a library; For example there are no `Semigroup/Monoid` instances for `Int` for that reason.
It is of course convenient to have _some_ pre-defined instances at hand that cover most of of the use cases, but ideally those would be provided by another (public sub-) library. Personally I don't make use of those instance more often than not, and pulling in `aeson` as a dependency is in my projects often unnecessary. So ultimately I'd like to get rid of the (mandatory) `aeson` dependency at some point.

As for this PR, I'm inclined to move the JSON-related instances for `OsPath` to an own module, say `OsPath.Aeson`, and add a flag `aeson-instances` to `path` that controls whether any JSON-related instances are build at all (defaults to `true`).

hasufell: > As for this PR, I'm inclined to move the JSON-related instances for `OsPath` to an own module, say `OsPath.Aeson`, and add a flag `aeson-instances` to `path` that controls whether any JSON-related instances are build at all (defaults to `true`).

cabal flags must not control exposed API though, because a package can not specify "I need this flag of one of my dependencies enabled". Cabal has no means to specify this, so the end user may specify an invalid configuration and get a compile error with missing modules.

src/Path/Include.hs
mmhat: Ah, ok, that's why it was solved this way! I will change that to use the newly added `OsPath.Internal.{Posix,Windows}.isWindows`.

validity-test-ospath/OsPath/Gen/Include.hs
mmhat: Yes, we did; But the one for `FilePath` is (obviously) the the same as the one for the general-purpose `String`, which is provided by the `genvalidity` package. Since there are no pre-defined instances for `OsPath`, `OsString` or `OsChar` and we have to roll our own, I thought that we might reconsider how we shrink paths specifically.
I didn't find the time to do that, but I didn't want to blindly copy existing code either; Hence the TODO...
But you are right, before we merge this I want to provide an implementation one way or the other.

mmhat: Revisiting the `genValid` implementation as well, this looks wrong to me as well: IMHO the generator should not only produce Unicode paths, but random sequences of `Word8`/`Word16` separated by `/`, `\` and `.`.
I.e. we should not assume an encoding.

validity-test-ospath/OsPath/Gen/Include.hs
mmhat: TBH, I simply don't know (yet) what constitutes a _valid_ `PosixPath`/`WindowsPath`... See also my comments in the previous thread.
@hasufell Do you have any idea what a sensible notion would be?

mmhat:
> @mmhat Great job on this work. It must have been quite a fiddly thing to get right, but you got `stack test --pedantic` to pass as-is on my machine the first time. 🎉

Thank you!

> I wasn't confident so I will just ask you: Can you please double-check that the original `Path` API isn't changed and no tests have been removed?

I am fairly certain that none got removed but more were added, but I will double-check that and provide a summary of what has changed :+1: 

Regarding our earlier discussion about `genvalidity` + `System.OsPath` instances: Ideally those would be provided by a `genvalidity-filpath` package as those are not specific to the `path` library. Would you accept the contribution of such a package in https://github.com/NorfairKing/validity?

src/Path/Include.hs
mmhat: BTW: Which code formatter is used by this project? I assume since Chris Done started it it's `hindent`?
Also, should I format the Cabal file with something like `cabal-fmt` or `cabal-gild` or not?

src/OsPath/Internal/Include.hs
mmhat: Ah, yes, you are right indeed. And frankly I should've known better before making such a proposal, given how much time I spent trying to build GHCup before I figured that I have to set a flag for `streamly`... I'm quiet happy that this changed.

validity-test-ospath/OsPath/Gen/Include.hs
hasufell: Writing a meaningf generator for windows paths is incredibly hard.

`filepath` has done this by expressing them in ABNF.

See here: https://github.com/haskell/filepath/blob/master/tests/filepath-equivalent-tests/Gen.hs

src/Path/Include.hs
NorfairKing: @mmhat Right, it gets compiled away anyway.
Right now there is no formatter for this project so don't worry about formatting it right now.

src/OsPath/Internal/Include.hs
NorfairKing: We've had a discussion about aeson before:
https://github.com/commercialhaskell/path/issues/136

If there is no good aeson instance (for example because it needs IO) then I'd rather not provide any.

NorfairKing:
> Regarding our earlier discussion about `genvalidity` + `System.OsPath` instances: Ideally those would be provided by a `genvalidity-filpath` package as those are not specific to the `path` library. Would you accept the contribution of such a package in https://github.com/NorfairKing/validity?

Absolutely!

validity-test-ospath/OsPath/Gen/Include.hs
hasufell: Depends what we mean with "valid". Do you mean the `isValid` function from filepath? It's a bit fuzzy. The only thing you need to keep in mind additionally for the new OsPath API when dealing with the constructors yourself is that windows paths are expected to be `[Word16]`. So if you create a one-byte `WindowsString` manually, almost all functions will throw `error`.

src/Path/Include.hs
mmhat: I changed that in the recent commits.

src/Path/Include.hs
mmhat: I removed that CPP in the recent commits.

validity-test-ospath/OsPath/Gen/Include.hs
mmhat: I think what I meant was answered here: https://stackoverflow.com/questions/1976007/what-characters-are-forbidden-in-windows-and-linux-directory-names

This is not a problem for the instance in the test-suite of this library since we do not pass the generated paths to any OS API, but if we move the instance to a more general-purpose service package like `genvalidity-filepath` we have to implement those restrictions.

validity-test-ospath/OsPath/Gen/Include.hs
mmhat: Uh, that looks horrifying! Thanks for the pointer though.

validity-test-ospath/OsPath/Gen/Include.hs
hasufell: As some of those SO answers say: it's almost impossible to have an exhaustive list of invalid filepaths on windows. The filepath `isValid` is just best effort.

validity-test-ospath/OsPath/Gen/Include.hs
hasufell: > Uh, that looks horrifying! Thanks for the pointer though.

It is beautiful, because it's a pretty good ABNF. I'm planning to actually represent filepaths through that data structure at some point. Otherwise it's not possible to have correct splitting etc. Lots of functions in filepath are ad-hoc.

NorfairKing:
@mmhat If you're ever blocked on me, please ping me.
AFAICT this is blocked on the TODOs, right?

mmhat:
@NorfairKing I started implementing `OsPath.Aeson`, but I got a bit lost there:
My idea was to provide functions and newtypes for `Path` such that downstream users can choose which representation they want to use for JSON. This led to quite a lot of definitions, see the module export list here: https://github.com/commercialhaskell/path/blob/5a16815182ed21093b2199f6dd068aecaffff023/src/OsPath/Aeson/Include.hs#L78
Since all of that requires converting between `PosixString`/`WindowsString` and JSON in numerous ways, which is IMHO outside the scope of this library, I moved those parts to a new project: https://github.com/mmhat/os-string-aeson
It is not ready yet, but I hope to push it to completion soon.
So my plan for the `aeson` instances is:

1. Complete and publish `os-string-aeson`, as it provides the foundation for the instances of the `OsPath.Path` type(s).
2. Complete and publish `path-aeson`, which provides the instances for those types. That library will depend on `os-string-aeson` and follow its design. 
3. Move the `aeson` instances for the `Path.Path` type(s) to `path-aeson` and close #136 .

Concerning this PR: I removed the `FromJSON`/`ToJSON` instances for the `OsPath` types, and as far as I can tell the only remaining TODO is the one here: https://github.com/commercialhaskell/path/blob/db0807f1ba06d091633947cecba423f05ab95acf/validity-test-ospath/OsPath/Gen/Include.hs#L161
In https://github.com/commercialhaskell/path/pull/192#discussion_r1653088349 I wrote that I don't really know what to put there, but right now I am inclined to say that a `PosixPath`/`WindowsPath` (from the point of view of this library) is indeed trivially valid.
This also matches notion of validity for the instances of the two types that are closest to an `OsPath`: `FilePath` and `ShortByteString`. The argument is roughly as follows:
For `Validity FilePath`, the documentation of `Validity [a]` states that a list is valid if all elements of the list are valid, and the documentation of `Validity Char` is trivially valid. Hence a `FilePath` trivially valid.
According to the documentation, a `ShortByteString` is trivially valid too, and since an `OsPath` can be seen as something between a `FilePath` and a `ShortByteString`, I suppose it is trivially valid as well.

NorfairKing:
> Move the aeson instances for the Path.Path type(s) to path-aeson and close
https://github.com/commercialhaskell/path/issues/136 .

This would break backward compatibility, which I'm not sure is a good idea.
We can put new instances in a separate package but leave the old ones in their current place for now.

> Concerning this PR: I removed the `FromJSON`/`ToJSON` instances for the `OsPath` types, and as far as I can tell the only remaining TODO is the one here:

You probably want to use an empty `instance Validity` then, not `validate = trivialValidation`. That way GHC fills in the right details using the default implementation.

validity-test-ospath/OsPath/Gen/Include.hs
```diff
@@ -157,5 +158,4 @@ instance GenValid PLATFORM_PATH where
       in
       shrinkedWithDrive <> shrinkedWithoutDrive
 
-instance Validity PLATFORM_PATH where
-    validate = trivialValidation -- TODO: Not yet implemented
+instance Validity PLATFORM_PATH
```
NorfairKing, line 161: :+1:

NorfairKing:
@mmhat what's the status of this?

mmhat:
@NorfairKing As far as I am concerned the implemetation is done. I still need to double-check that the test coverage of the  `test` testsuite didn't change, but I plan to do that at the beginning of the next week.

mmhat:
**TLDR;** I double-checked that no tests were lost in the refactoring of the `test` testsuite; None were lost indeed and in fact the test coverage is slightly better.

Here's a short summary:

 * The parsing tests remained the same, only the helper funtions were moved to the `Common/Inlude.hs`.

 * The TH tests went to `TH.{Posix,Windows}` modules.

 * The tests for the operations on `Path`s were moved to the `Common/Include.hs` file, since they are/should be independent of the platform,
   e.g. for "parent (directory </> filename) == directory" it must not matter whether we are on Posix or Windows.

 * The biggest change is that absolute paths (on Windows) are tested in a more principled way:
   Before there were some tests that came with flavours for "C:\", UNC and DOS Device drives while others just tested one of those (the first one in the list).
   For example, previously we had a test "dirname C:\ must be a Rel path", now we also test that for the other drives as well.

   This is done by defining a `drives_ :: NonEmpty FilePath` in `Common.{Posix,Windows}` which hold lists of the possible drives for the respective platform, and a helper function `forDrives :: (Path Abs Dir -> Spec) -> Spec` in `Common/Include.hs` is used to iterate over those lists.

 * Likewise, the extension tests test all valid extensions.
   The additional tests are IMHO not terribly important, but since they're cheap they don't hurt either.

@NorfairKing Let me know if you have questions about the changes or if I need to change something; As far as I am concerned this can be merged.

NorfairKing:
@mmhat The tests pass locally but I'm having a hard time trying it out with other projects because `path` now has dependencies that aren't in a released version of `nixos` yet and are newer than the latest LTS that's in the latest nixos release.
(I.e. nixos-24.05 and lts-22.22.)
I'm guessing there's no way around that because that's where OSPath coomes from. (Is that right?)
But that also means I can't try out this new version of `path` for backward compatibility, and breaking backward compatibility isn't an option IMO.

One option is to wait for `filepath-1.5` to be common (in an lts that's in a stable release of nixos), but I don't like that option very much either.

Do you have any ideas?

NorfairKing:
I just manaaged to get some compilation happening and I see these:

```
path                  > <no location info>: error: [GHC-42258] [-Wunused-packages, Werror=unused-packages]
path                  >     The following packages were specified via -package or -package-id flags,
path                  >     but were not needed for compilation:
path                  >       - os-string-2.0.2.2 (exposed by flag -package-id os-string-2.0.2.2-IgJUWyRUhjdET05tAskRpx)
```
Have you turned on `Wunused-packages`?

```
path                  > /home/syd/src/path/src/OsPath/Internal/Include.hs:58:1: error:
path                  >     Ambiguous module name ‘System.OsString.Posix’:
path                  >       it was found in multiple packages:
path                  >       filepath-1.4.300.1 os-string-2.0.2.2
path                  >    |
path                  > 58 | import qualified System.OsString.PLATFORM_NAME as OsString
path                  >    | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```

mmhat:
@NorfairKing Well, there is a way to relax the version bounds of filepath/os-string to include `filepath >=1.4.100.0` as described [here](https://hackage.haskell.org/package/os-string-1.0.0).
The problem is that the `System.OsString` modules shipped with `filepath >=1.4.100.0 && <1.5.0.0` do not include two functions I use (`all` and `null`); That's the reason for the rather high version bounds.

I am working on (internal) compatibility modules to work around that right now.

mmhat:
> The problem is that the `System.OsString` modules shipped with `filepath >=1.4.100.0 && <1.5.0.0` do not include two functions I use (`all` and `null`)

Turns out there were some more functions missing; See [compat/System/OsString/Compat/Include.hs](https://github.com/commercialhaskell/path/blob/b2d8b954ed3e0b861e8debfb2fe6e47e8eff8201/compat/System/OsString/Compat/Include.hs).

@NorfairKing I added a flag `old-os-string` that controls whether `filepath >=1.4.100.0 && <1.5.0.0, os-string <2.0.0.0` or `filepath >=1.5.0.0, os-string >=2.0.0.0` are pulled in as dependencies. Can you check again if that works for you too? (Since I am not a Nix user I am particularly interested if it works with that setup.)

NorfairKing:
@mmhat now I'm seeing this:

```
       In the dependencies for hashable-1.4.4.0:
         * os-string must match >=2.0.2, but os-string-1.0.0 is in the Stack configuration (latest matching version is 2.0.6).
       The above is/are needed due to path-0.9.5 -> hashable-1.4.4.0
```

We're getting closer though!

path.cabal
```diff
@@ -74,21 +74,15 @@ library
...
 
-  if flag(old-os-string)
-    build-depends:   filepath   >= 1.4.100.0
-                   , os-string  < 2.0.0
-  else
-    build-depends:   filepath   >= 1.5.0.0
-                   , os-string  >= 2.0.0
-
-library compat
+library os-string-compat
```
NorfairKing, line 82: Are you sure this will be stack-compatible? That's a must.

mmhat:
> ```
> extra-deps:
> - os-tring-1.0.0@sha256:f1582da15c52761681da87d98c854a0db54049f34d59f74cfec4cc56ad8fbc05,1852
> flags:
>   path:
>     old-os-string: true
> ```

There's a typo : `os-tring` should probably be `os-string`... Not sure if that's the source of your problems though.

Which resolver are you using?
I had troubles with the stack configuration yesterday as well: The reason was that the latest LTS snapshot has `filepath <1.5` but `os-string >=2`, and in addition to that the hashable version in that snapshot has also conditional dependencies on both libraries (See the `impl(ghc >= 9.2)` in the [package description](https://hackage.haskell.org/package/hashable-1.4.4.0/hashable.cabal).)

path.cabal
mmhat: Yes, it is; Stack supports depending on (public) sublibraries (of other packages) since 2.15.1; Not sure since which version depending on private sub-libraries is supported...

NorfairKing:
> > ```
> > extra-deps:
> > - os-tring-1.0.0@sha256:f1582da15c52761681da87d98c854a0db54049f34d59f74cfec4cc56ad8fbc05,1852
> > flags:
> >   path:
> >     old-os-string: true
> > ```
> 
> There's a typo : `os-tring` should probably be `os-string`... Not sure if that's the source of your problems though.

Ah that was a copy-paste error, I had it right in my source code.

> Which resolver are you using? I had troubles with the stack configuration yesterday as well: The reason was that the latest LTS snapshot has `filepath <1.5` but `os-string >=2`, and in addition to that the hashable version in that snapshot has also conditional dependencies on both libraries (See the `impl(ghc >= 9.2)` in the [package description](https://hackage.haskell.org/package/hashable-1.4.4.0/hashable.cabal).)

lts-22.22, the latest stable release of nixos.
Ideally all of this would "just work" without any flags needing to be set.

mmhat:
@NorfairKing I added a `stack.old.yaml` that builds with the `old-os-string` flag enabled. Hope that helps...

> Ideally all of this would "just work" without any flags needing to be set.

Yeah, that would be great. I am not sure why one has to set flags explicitly, but I suspect that Stack solves dependencies without trying to toggle flags? TBH, I have no idea how to work around this issue...

NorfairKing:
@mmhat for least annoyance, we need to make sure that `path` "just works" (with default flags) with at least stackage lts-22.22 (the current stable release) and lts-22.31 (the next release, so far).
If we can't do that then we need to look upstream to make them forward compatible with this future release before we release it.
We can't just throw code over the wall and expect it to stick.

Right now I still see this when trying to upgrade only path on a 22.22 project:

```
       In the dependencies for path-0.9.5:
         * filepath must match >=1.5, but filepath-1.4.300.1 is in the Stack configuration (latest matching
           version is 1.5.3.0).
       The above is/are needed since path is a build target.
```

Here's another project:

```
       In the dependencies for path-0.9.5:
         * filepath must match >=1.5, but filepath-1.4.300.1 is in the Stack configuration (latest matching version is 1.5.3.0).
       The above is/are needed since path is a build target.
```

You can try it out by adding `/home/youruserhere/src/path` to the `packages` section of the `stack.yaml` of https://github.com/NorfairKing/smos on `development`.

If we can't make this backward compatible then we may have to make this PR should a separate package so that we definitely don't break anyone downstream.
(There is a LOT downstream of this package.)

mmhat:
@NorfairKing I managed to make the changes backwards compatible to a certain degree; I think this as good as it can get.
The current LTS builds with no additional `extra-deps:` and no `flags:`; Just specifiying a `resolver:` is sufficient.
(See the checked in stack.yaml)

> You can try it out by adding `/home/youruserhere/src/path` to the `packages` section of the `stack.yaml` of https://github.com/NorfairKing/smos on `development`.

Adding the following to `extra-deps:` yields a successful `stack build smos` on both the release and the development branch of your project:
```yaml
- Cabal-3.10.3.0@sha256:4a36954a66769eef90353daa0dbb07d43e49c228281383c7e4134445d495f714,12313
- Cabal-syntax-3.10.3.0@sha256:c2ca36499bf9365726968b1e11a757e6d846c47cdf0c2c26a003698fd3c300fe,7431
- directory-1.3.8.5@sha256:fbeec9ec346e5272167f63dcb86af513b457a7b9fc36dc818e4c7b81608d612b,3166
- filepath-1.4.300.2@sha256:345cbb1afe414a09e47737e4d14cbd51891a734e67c0ef3d77a1439518bb81e8,5900
- process-1.6.21.0@sha256:685bc68759da31b5f152092fe664e1644e84f6dc0ae7a6c143e8564a1d6dafe8,2644
- unix-2.8.5.1@sha256:3f702a252a313a7bcb56e3908a14e7f9f1b40e41b7bdc8ae8a9605a1a8686f06,9808
```

There is unfortunately no way to drop the `filepath >=1.4.300` constraint; But I think it's not too much to ask to add some packages to `extra-deps:` if one want's to use a snapshot `< 22.21` together with a recent upstream package. 

> 
> If we can't make this backward compatible then we may have to make this PR should a separate package so that we definitely don't break anyone downstream. (There is a LOT downstream of this package.)

Yeah, I know; I myself use `path` extensively...

Is there anything left I can do to improve this PR?

NorfairKing:
> Is there anything left I can do to improve this PR?

I'm replying asap to that you know I care about getting this merged, but can't review tonight.
I'll see what I can do tomorrow.

path.cabal
```diff
@@ -1,95 +1,193 @@
...
 flag dev
   description:        Turn on development settings.
   manual:             True
   default:            False
 
+flag os-string
+  description:        Use an older version of the os-string library.
+  manual:             False
+  default:            False
+
```
NorfairKing, line 35: This doesn't make sense to me.
The default is `False`, but if you turn on the flag it becomes .. `False`?

path.cabal
```diff
@@ -1,95 +1,193 @@
...
+  exposed-modules:   System.OsString.Compat.Posix
+                   , System.OsString.Compat.Windows
+
+  build-depends:     base       >= 4.12    && < 5
+
+  if flag(os-string)
+    build-depends:   os-string  >= 2.0.0
+  else
+    build-depends:   filepath   >= 1.4.100.0
+    --               , os-string  < 2.0.0
```
NorfairKing, line 101: This comment can be removed now.

NorfairKing:
@mmhat I left two more little comments. After that I think it's ready to merge.
I'll take care of that once this is ready.

path.cabal
NorfairKing: Oh I think I completely misunderstood what `manual` means. Then you can ignore this comment.

mmhat: Ok, good! Referencing the relevant documentation here anyway: https://cabal.readthedocs.io/en/stable/cabal-package-description-file.html#pkg-field-flag-manual

path.cabal
mmhat: Done.

mmhat:
@NorfairKing Ok, this can be merged. Thank you very much for maintaining this library!

NorfairKing:
Merged and released as `0.9.6`.

NorfairKing:
I can't tell why this is not being marked as merged so I will close it manually.

PR #194, opened 2024-09-11
Improved `Lift` instances for Path/OsPath types

mmhat:
Extend the `Lift` instances such that they support parametric types and promoted data constructors for the two parameters of the Path/OsPath types.

### DIFF
--- a/path.cabal
+++ b/path.cabal
@@ -69,6 +69,8 @@ library
                    , OsPath.Internal.Posix
                    , OsPath.Internal.Windows
 
+  other-modules:     Utils
+
   build-depends:     aeson      >= 1.0.0.0
                    , base       >= 4.12    && < 5
                    , deepseq
--- a/src/OsPath/Internal/Include.hs
+++ b/src/OsPath/Internal/Include.hs
@@ -54,6 +54,7 @@ import qualified System.OsPath.PLATFORM_NAME as OsPath
 
 import System.OsString.Compat.PLATFORM_NAME (PLATFORM_STRING)
 import qualified System.OsString.Compat.PLATFORM_NAME as OsString
+import Utils (typeableToType)
 
 -- | Path of some base and type.
 --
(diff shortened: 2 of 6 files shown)


NorfairKing:
The `Path` type is not meant to be used with anything other than `Abs`, `Rel`, `File`, `Dir`, see https://github.com/commercialhaskell/path?tab=readme-ov-file#the-data-types

> (Why not use data kinds like data Type = File | Dir? Because that imposes an extension overhead of adding {-# LANGUAGE DataKinds #-} to every module you might want to write out a path type in. Given that one cannot construct paths of types other than these, via the operations in the module, it’s not a concern for me.)

Could you elaborate on your aims here?
As-is, I'm very hesitant to consider merging this.

mmhat:
@NorfairKing Well, I want to (re-)use the `Path` type with a custom base :slightly_smiling_face: I missed that part of the documentation though.

I have some application code that creates symlinks to filesystem entries in parent directories, so I construct a path containing `..`s and stored that in a `Path (WithParents b) Dir` value. Recently I wanted to extract that code to an own library (I do not intend to release that on Hackage.) and when I wanted to add a quasi-quoter I got type errors since the `Lift (Path b t)` annotates the splice with the wrong type. Hence I opened this PR.

While I understand (and agree) that `path` does not aim to support use cases like that in general, would it be possible to only modify the `Lift` instance since it is the one thing one cannot work around? If the complexity added by this PR is deemed to much, would you be open to a change limiting the defined instances to:
```haskell
instance Lift (Path Abs Dir)
instance Lift (Path Abs File)
instance Lift (Path Rel Dir)
instance Lift (Path Rel File)
```

NorfairKing:
> While I understand (and agree) that `path` does not aim to support use cases like that in general, would it be possible to only modify the `Lift` instance since it is the one thing one cannot work around? If the complexity added by this PR is deemed to much, would you be open to a change limiting the defined instances to:

I want to say yes but I find it difficult to estimate what kind of breakage that would cause.
Do you have any more insight into this?

NorfairKing:
Closing this due to lack of activity. Feel free to reopen.

